PR #40, opened 2015-09-08
added proper lts chart, added note about 4.4.1 being an example

rvagg:


thefourtheye:
The image is still attributed to nodesource?

rvagg:
yes

thefourtheye:
@rvagg I was thinking that we would be using "Node.js Foundation" everywhere. Where should we use Nodesource?

rvagg:
this was designed by a NodeSource designer and is therefore (c) NodeSource, but it's licensed under creative commons

thefourtheye:
Cool. LGTM.

PR #73, opened 2016-01-29
website: have website point to lts_schedule #493

eljefedelrodeodeljefe:
as per discussion on nodejs/nodejs.org#493 and in reference to nodejs/nodejs.org#494. We would want to have a link to the picture of the schedule in the readme, instead of a link to the raw .png.

This would be much appreciated.


eljefedelrodeodeljefe:
Thanks.

PR #100, opened 2016-05-02
Correct v0.12 LTS End Date

RochesterinNYC:
According to the release and support diagram, it seems that `v0.12` of node is slated to end LTS on April 1st, 2016. The table displayed however says that LTS for `v0.12` ends on April 1st, 2017. I think is an error?

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ beyond the LTS release.
   <td>v0.12</td>
   <td>(current)</td>
   <td>2016-04-01</td>
-  <td>2017-04-01</td>
+  <td>2016-04-01</td>
 </tr>
 <tr>
   <td>v4.2.0</td>


RochesterinNYC:
Either that, or the diagram is incorrect? Unsure now at this point.

TheAlphaNerd:
There are two different dates here being conflated... The diagram shows `Active-LTS` in blue and `Maintenance LTS` in grey (shortened to simply Maintenance).

The LTS end date is the end to the entire support cycle including Maintenance.

What is confusing is that the diagram show v0.12 ending some time after Dec 2016 but ending before april 2017... which is not inline with the documentation above. I'm going to leave this open so people can chime in and let you know if that date should be changed, the current change should not be landed though.

Thanks for taking the time to double check this stuff, it is important and we appreciate the help!

RochesterinNYC:
Okay, so am I correct in understanding that the diagram is a little inaccurate and the Maintenance (LTS) bar for Node.js `v0.12` should actually extend to Apr 2017? Thanks for your clarifications!

Also, thanks for clarifying on the Maintenance actually being Maintenance LTS point. That was a little confusing at first.

TheAlphaNerd:
@RochesterinNYC both are accurate-ish. There is some wiggle room there based on usage and need.

v0.10 and v0.12 are special cases that fall outside the normal process.

/cc @mikeal

rvagg:
That table should probably read `2016-12-31` because that's when we've agreed to can 0.10 because that's when OpenSSL 1.0.1 support ends.

On that topic, however, I did chat to one of the OpenSSL guys about this and they were pretty firm on ceasing _their_ support for it but did mention that some of the OS vendors are talking about what they might do (because they have LTS schedules that go a lot further than the end of 2016) so there's a chance that support for 1.0.1 might be picked up by another party and perhaps there's something we can piggy-back to extend our support a little.

Not that I really want 0.12 to survive too much longer tho!

PR #105, opened 2016-05-16
Losslessly optimize schedule.png by zopflipng

PeterDaveHello:
Size change: 117739 (114K) -> 95588 (93K), 81.186% of original.


TheAlphaNerd:
Do we need to keep it at the current resolution as well? Would it be useful to keep the original in the repo too uncompressed but serve the compressed version in the readme?

jasnell:
I'm good with this. I don't see much reason to keep the original. LGTM

TheAlphaNerd:
kk... LGTM

PeterDaveHello:
Resolutions of these two pics are the same, git repo has copy of the old one, so there is no need to keep them both, and it is totally compatible with origin file.

TheAlphaNerd:
closing as we have a new image... please feel free to make a new PR

PeterDaveHello:
why not just update it in the same pr?

PR #121, opened 2016-07-12
README redo

rvagg:
While I've taken many elements from the current README, this is basically a rewrite.

The main things I'm proposing here are:
- Use this as a place to document our "Release and LTS Plan", i.e. pretty much everything about release policies spanning Current to Maintenance. It could go even further but this is a start. I did this because there's so much in here that touches on non-LTS which makes it's difficult to talk about LTS in isolation. Plus, we don't have good documentation for this stuff anywhere else. So why not unify it and manage it here? We point people to this page when talking about the differences in releases anyway so we should probably have more than just strictly LTS.
- Getting that mid-current V8 upgrade policy finally documented
- Update of the main graphic with a cleanup and to deal with the 2 months after "Current" for non-LTS lines. I've marked them in grey but not given that period a name. The colour is the same as for "Maintenance" because we discovered from our experience with v5 that we wanted to treat it like that, so let's just roll with it eh?
- Introduced an explainer version of the graphic that is limited to v4+ and is animated to show 3 major points of the release/LTS plan we have. I believe this should be enough to get the basic message across about how we do things. I try not to jam in too much detail into the graphic but expand further on it in the rest of the document. It's possible that there needs to be some more text in the "Summary" section to make it a more complete summary but I'll need some feedback on that.
- Put a license on the doc, I copied what Inclusivity did at the bottom of theirs
- Put a list of LTS WG members at the bottom—this comes from the @nodejs/lts team and it's probably time to review that list and ask those folks who haven't been active if they'd like to be removed from the list.

I'd like to get some discussion amongst @nodejs/lts first before taking this to Collaborators and then CTC for sign-off.

FYI I did sentence-linebreaking in this doc as a compromise between char limit line breaks (which I hate) and full paragraph linebreaks which I prefer but are a bit more difficult to review in PR form. Feedback on that is welcome too of course!


README.md
```diff
@@ -1,112 +1,204 @@
...
+  <td align="center">2015-10-01</td>
+  <td align="center">2017-04-01</td>
+  <td align="center">2018-04-01</td>
 </tr>
+
 <tr>
-  <td>N/A</td>
-  <td>v5</td>
+  <th>v5</th>
+  <td>Current&nbsp;(non-LTS)</td>
```
Fishrock123: I still think this should be `"Maintenance" (non-LTS)`

README.md
```diff
@@ -1,112 +1,204 @@
...
+
+## Support timeframes
+
+Every release line covered by LTS will be actively maintained for an initial period of 18 months from the date it enters LTS coverage; this is "**Active LTS**".
+Following those 18 months of active support, the release line will transition into "**Maintenance**" mode for 12 additional months.
+
+Given this schedule, there will be no more than two "Active LTS" release lines at any given time, overlapping for a maximum period of six months.
+Users are encouraged to use this overlap to plan a migration path between LTS release lines.
+
+Release lines not covered by LTS (i.e. odd-numbered release lines) receive an additional 2 months of "Maintenance"-style support following their 6 months as "Current" releases.
```
Fishrock123: remove `-style`, imo. "Maintenance" with quotes should do.

README.md
```diff
@@ -1,112 +1,204 @@
...
+
+## Acceptable changes for release lines
+
+All release lines covering a single semver-major version number, be it labelled "Current", "Active LTS" or "Maintenance" will:
```
Fishrock123: Maintenance branches should never receive minors except for major security issues.

README.md
```diff
@@ -1,112 +1,204 @@
...
+
+Release lines in "Active LTS" and "Maintenance" will be pinned to a single major version of V8 during their lifetime in order to maximise stability. Bug fixes and improvements may be introduced, either by downstreaming, backporting, or by original authorship, according to the same policy for Node.js changes for LTS release lines.
+
+Release lines in "Current" _may_ receive updates to their V8 dependency. This will occur ***only if*** it can be done in a way that is not semver-major and does not break the stable ABI of the current major version. This may be achieved by either receiving an ABI-stable update from upstream or by applying floating patches to the dependency to ensure ABI-stability. Implementation details are left up to the Node.js V8 team.
+
+Upgrades to V8 in a "Current" release line should not occur within two months of that release line moving in to "Active LTS". i.e. upgrades to V8 should not occur from August through to the transition of "Current" into LTS in October.
+
+### Changes acceptable for "Active LTS"
```
Fishrock123: Should also have some sort if definition of minors we backport, maybe?

Fishrock123:
I think we should define ABI.

bnoordhuis:
It links to the Wikipedia article.  Are you thinking of something more specific?

README.md
```diff
@@ -1,112 +1,204 @@
...
+
+Release lines in "Active LTS" and "Maintenance" will be pinned to a single major version of V8 during their lifetime in order to maximise stability. Bug fixes and improvements may be introduced, either by downstreaming, backporting, or by original authorship, according to the same policy for Node.js changes for LTS release lines.
+
+Release lines in "Current" _may_ receive updates to their V8 dependency. This will occur ***only if*** it can be done in a way that is not semver-major and does not break the stable ABI of the current major version. This may be achieved by either receiving an ABI-stable update from upstream or by applying floating patches to the dependency to ensure ABI-stability. Implementation details are left up to the Node.js V8 team.
+
+Upgrades to V8 in a "Current" release line should not occur within two months of that release line moving in to "Active LTS". i.e. upgrades to V8 should not occur from August through to the transition of "Current" into LTS in October.
```
ofrobots: This is somewhat ambiguous because August through to October would be three months.

README.md
```diff
@@ -1,112 +1,204 @@
...
-  <td>2019-04-01</td>
+  <th>v6</th>
+  <td>Current&nbsp;(pending&nbsp;LTS)</td>
+  <td align="center">2016-10-01</td>
+  <td align="center">2018-04-01</td>
+  <td align="center">2019-04-01</td>
+</tr>
+
+<tr>
+  <th>v7</th>
```
richardlau: Mark v7 as `(non-LTS)`?

README.md
rvagg: I'm OK with this but I was being vague about it because so far we've resorted to leaving a label off v5 after we got v6. Having it explicitly called "Maintenance" would probably be a step up from that and clarify how it works.

README.md
rvagg: done

README.md
rvagg: goes with my comment above, I think we should have consensus in the LTS WG that we want to adopt this wording (it's not strictly an LTS concern but we seem to be the ones most concerned with release management).

README.md
rvagg: @ofrobots actually I was thinking of expanding on the ambiguity and leaving it up to both the LTS WG and the nodejs/v8 team to negotiate on exact timing rather than being strict about it in this document. Since we may not get an LTS till late October then perhaps having a late August V8 upgrade isn't a big deal. It also comes down to the nature of the V8 upgrade at that time, perhaps it's very minor or perhaps it does something big like replace the GC or something that changes the performance profile in a major way or introduces new risks that we're concerned about.

Would you be OK with introducing explicit flexibility in here and make it guidance rather than a rule?

README.md
rvagg: Yeah, but I couldn't come up with any! So I left it above as:
  - sign-off from at least one active member of the LTS Working Group;
  - no objection from any active member of the LTS Working Group

Meaning that our reasoning can evolve over time and perhaps we can codify it when we end up with an established pattern; I just don't see that now in how we're working and don't want to prematurely make things strict. See https://github.com/nodejs/node/issues/7714 (backport `fs.mkdtemp()`) for an example of this, I'm not comfortable ruling it in or out because I don't see a clear stance we're prepared to make as a group.

Unless you want to propose some wording that gets us in the right direction?

rvagg:
I think I agree with @Fishrock123 on defining ABI, or at least expanding it. It's a foreign concept for most Node users and I find myself always explaining it (or being corrected by people: "API you mean?"). I'll see if I can put a sentence or two along with the link for further info.

shellberg:
I'd like to see some definition for the scope (or criteria) of backporting fixes.  So far, the text tends to reference the downstreaming controlled by the classification of semver.[major | minor] only, and in practice by the use of the 'lts-watch' tags, for example.
Please consider: if an issue is raised from a 'Maintenance' release, i.e. one that has (perhaps just) passed from LTS into Maintenance, it seems there is no commitment currently that would consider addressing this in such a way that any fix (developed on master) is backported beyond the LTS releases to actually address the Maintenance release on which the issue was raised?  Current practice seems to be only to consider ports of fixes from Master to LTS, I think from the limited cases I've been able to follow since adopting the LTS regime at 4.X.
The specific policy of limiting only Critical classified issues would seem to potentially blocking delivering the fix for a functional problem to a 'Maintenance' release stream on which that problem may have originally been reported.

Another aspect to potentially consider is also how we are ensuring the quality of stability of the older releases.  There is discussion of the structure of the LTS branches (staging and release).  However, as far as I can determine, any changes to be delivered into any release are always received direct from the (unstable) Master stream?  Meaning, before cutting a new Release (LTS or Maintenance) we are currently receiving fixes/changes that, subject to code review and blessing of an LTS workgroup member, may be only days (or hours) old that haven't had any timescale in which they have demonstrated stability (even in Master!)?  Should there be a process defined where a change/fix being backported is allowed to 'settle' in a 'Current' release streams delaying the application to the 'mostly stable' LTS, and 'most-stable' Maintenance releases?  Is this effectively what is happening under the current LTS regime somehow? Or by only considering Critical fixes (cf Security), that effectively no other fixes are/will ever result in a changed Maintenance release.
(Secondarily, should a change destined for a 'stable' (cf 'Maintenance') release always be sourced only from Master, and not an intervening (LTS) release based on the sliding designation of stability where and once its been integrated and proven?  There may be experience here by those who routinely integrate fixes/changes into former releases... is the fix in Master always the best basis?)
In other words, is there a timing dimension that could be introduced in the backport process apart from just the criticality of the change that could help to demonstrate quality stability?

README.md
mhdawson: Or just change the wording to "For example".  I think just saying should not within 2 months is reasonable as the intention and then as everything there can be exceptions. Then change the last sentence to "For example, assuming LTS ships at the end of October upgrades should not occur between then end of August and that time."

mhdawson:
My understanding is that we do have a "settling" time in the "Current" release before PR's are backported to LTS.  I think adding some text to describe that would make sense.

README.md
ofrobots: I think it would be good to state it as an intent but allow for some flexibility as deemed appropriate by the LTS and V8 teams. I didn't parse the first sentence as offering much flexibility, and the second sentence ambiguously expands it to three months.

README.md
Fishrock123: Well, it's no Longer current so maybe literally call it 
`¯\\\_(ツ)\_/¯`

Fishrock123: We *could* call it just `(non-LTS)` I guess. I'd really prefer not to though.

README.md
jasnell: I'm fine with the wording on this. It gets the point across.

README.md
```diff
@@ -1,112 +1,204 @@
...
+* Receive commits labelled "[***semver-minor***](https://github.com/nodejs/node/pulls?utf8=%E2%9C%93&q=is%3Apr%20label%3Asemver-minor%20)".
+  However, it is expected that release lines that are covered by LTS shall receive a minimal number of these changes in order to maximise stability.
+  Additionally, semver-minor changes will be delayed and grouped as much as is practical in LTS release lines in order to ease the upgrade path for users of that line.
+  Any semver-minor commits proposed for inclusion in an LTS release line will require:
+  - sign-off from at least one active member of the LTS Working Group;
+  - no objection from any active member of the LTS Working Group
+* Receive semver-patch commits, i.e. those not labelled either semver-major or semver-minor as appropriate.
+  Any semver-patch commits proposed for inclusion in an LTS release line will require:
+  - sign-off from at least one active member of the LTS Working Group;
+  - no objection from any active member of the LTS Working Group
```
jasnell: I'm wondering if we should include an explicit mention of CITGM here? No commit should land in an LTS that breaks anything in a CITGM run.

README.md
```diff
@@ -1,112 +1,204 @@
...
-period of 18 months from the date it enters LTS coverage. Following those 18 
-months of active support, the major version will transition into "maintenance" 
-mode for 12 additional months.
+It is possible that ***critical*** security and bug fixes may be labelled "semver-major".
+Such situations are expected to be rare.
 
-Given this schedule, there will be no more than two active LTS releases at any 
-given time, overlapping for a maximum period of six months.
+Where it is necessary to introduce such "semver-major" changes into a release line covered by LTS.
+The changes will result in an increment of the semver-minor version number only.
```
jasnell: May want to add a short explanation about *why* semver-major bumps in LTS streams are handled as semver-minors... specifically, the Semantic Versioning spec does not provide a mechanism for us to adequately communicate that a semver-major change has occurred within an LTS release line.

jasnell:
Few comments, but overall LGTM

README.md
mhdawson: At this point I'm ok with landing as is and considering adding a CITGM reference later on.

jasnell:
Closing this due to lack of forward progress on it. Can reopen if necessary

PR #140, opened 2016-10-02
doc: fixing v0.10 drop support date

italoacasas:
To avoid more misunderstandings, I feel we should change the `end date for v0.10` and display the correct date.

cc: @jasnell


TheAlphaNerd:
lgtm

italoacasas:
sorry I forgot about `v6` dates.

README.md
```diff
@@ -39,9 +39,9 @@
 <tr>
   <td><b>Pending</b></td>
   <td>v6</td>
-  <td>2016-10-01</td>
-  <td>2018-04-01</td>
-  <td>2019-04-01</td>
+  <td>2016-10-25</td>
```
TheAlphaNerd: this should be the 18th

jasnell:
Should be the 25th tentatively right? It's supposed to coincide with the
release.

italoacasas:
@jasnell I think in the last CTC meeting that you guys did, someone makes the point of release LTS first because something (I don't remember why right now).

TheAlphaNerd:
In the CTC meeting last week we discussed doing the LTS release a week earlier if possible to simplify things. The logic was the less moving parts the less chance for error... it would also give us a week to focus exclusively on v7.

If you think it would be better to have them on the same date I don't see why we can't. We can add this to the agenda of tomorrow's meeting and I can proxy your opinion

jasnell:
That works for me!

TheAlphaNerd:
landed in 3821a26...75bd36a

PR #141, opened 2016-10-03
meta: updated messaging regarding dates

MylesBorins:
Currently our schedule specifies exact dates, this can be hard to
commit to. If we plan to be flexible with the dates we chose when
moving a release stream between different levels of support we should
be explicit about it.

This removes all future dates that have currently been decided,
alternatively offering a month in which the event will take place.

It also adds some new language the outlines when individuals should
expect us to give them a date.
- No later than the first of the month the change is happening
- No less than 14 days before the change is happening

I believe that this contract is reasonable and much less likely
to cause miscommunication in the future.

One thing not mentioned in here is how we will be communicating these
dates, and potential changes. We need to decide on a single communication
channel and stick to it. Very likely it can be the blog, but we should
discuss this first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,15 +4,15 @@
 
 | Release |  LTS Status   | Codename | Active LTS Start | Maintenance Start | Maintenance End |
 |   :--:  |    :---:      |   :---:  |       :---:      |       :---:       |      :---:      |
-|  v0.10  |**End-of-Life**|    -     |        -         |    2015-10-01     |   2016-10-31    |
-|  v0.12  |**End-of-Life**|    -     |        -         |    2016-04-01     |   2016-12-31    |
-| [4.x][] |**Maintenance**| [Argon][]|    2015-10-01    |    2017-04-01     |   2018-04-01    |
+|  v0.10.x  |**End-of-Life**|    -     |        -         |    2015-10-01     |   2016-10-31    |
+|  v0.12.x  |**End-of-Life**|    -     |        -         |    2016-04-01     |   2016-12-31    |
+| [4.x][] |**Maintenance**| [Argon][]|    2015-10-01    |    2017-04-01     |   April 2018    |
 |  5.x    |No LTS         |          |                  |                   |                 |
-| [6.x][] |**Active**     | [Boron][]|    2016-10-18    |    2018-04-18     |   2019-04-18    |
+| [6.x][] |**Active**     | [Boron][]|    2016-10-18    |    April 2018     |   April 2019    |
 |  7.x    |No LTS         |          |                  |                   |                 |
-|  8.x    |**Pending**    | Pending  |    2017-10-01    |    2019-04-01     |   2019-12-31<sup>2</sup> |
+|  8.x    |**Pending**    | Pending  |    October 2017  |    April 2019     |   December 2019<sup>2</sup> |
 |  9.x    |No LTS         |          |                  |                   |                 |
-| 10.x    |**Pending**    | Pending  |    2018-10-01    |    2020-04-01     |   2021-04-01    |
+| 10.x    |**Pending**    | Pending  |    October 2018  |    April 2020     |   April 2021    |
 
 * <sup>1</sup>: All scheduled dates are subject to change by the Node.js LTS
   working group or Node.js Core Technical Committee.
@@ -39,6 +39,10 @@ Every major version covered by the LTS plan will be actively maintained for a
 period of 18 months from the date it enters LTS coverage. Following those 18
 months of active support, the major version will transition into "maintenance"
 mode for 12 additional months.
+
+The exact date that a release stream will be moved to LTS, moved between LTS
+modes, or deprecated will will be chosen no later than the first day of the month.
+If it is to be changed, it will be done with no less than 14 days notice.
 
 Given this schedule, there will be no more than two active LTS releases at any
 given time, overlapping for a maximum period of six months.


nathanhammond:
This seems fine. Maybe some wordsmithing still to make it clear, but I know the intent because of our [conversation](https://twitter.com/nathanhammond/status/782811059295952896).

That being said, dropping of support on specific dates seems _less_ constraining than specifying start dates. They will always simply be arbitrary lines in the sand. I'm unsure why we can't simply select one.

The only reason to make them coincide with releases would be to make it possible for migration to a new version–but nobody should reasonably attempt to move from a fully EOL LTS to whatever Node.js version is being contemporaneously released as that's a bad technical decision. Even if you said that it was to trail the new release by N days it's a bad technical decision because you don't know what issues you may encounter and you shouldn't expect to be able to solve them in 7-14 days. (Thus the months-long overlap periods.)

nathanhammond:
Hrm. There are three transitions that matter:

1. Current to Active.
1. Active to Maintenance.
1. Maintenance to EOL.

My previous comment only addresses point 3. This language is certainly needed for the other transitions (which simply aren't relevant with [our approach in the Ember community](http://emberjs.com/blog/2016/09/07/ember-node-lts-support.html)).

So I propose that final EOL has a date attached and transitions 1 and 2 follow the pattern proposed by this PR.

TheAlphaNerd:
@nathanhammond I'm going to have to think about this for a bit. If I am following you correctly you are saying that the only date that should be set, and set FAR in the future is the EOL date, which should be decided when a release goes from Current to Active.

At this current time my gut is telling me that this date needs to be flexible as well. Here is a quick breakdown of the security releases we had to do last week.

* Thu 22 [we are notified of CVE's](https://www.openssl.org/news/secadv/20160922.txt)
* We plan for a release on the 27th,  including the openssl CVE as well as [some other vulnerabilities that we are patching](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V010.md#notable-changes)
* On Monday the 26th [a more severe CVE is released, caused by the last update](https://www.openssl.org/news/secadv/20160926.txt)
  - this did not affect the 1.0.1 line that 0.10 is on, but these things happen
  - these were fixed before we even did a release, no Node release was never vulnerable

I don't feel that it would be particularly responsible to EOL v0.10 less than a week after these changes have landed. It is important to have this kind of flexibility, and as such it seems that choosing a date, that is subject to change, gives the wrong message.

Is a date range of a month to broad for your needs?

nathanhammond:
> Is a date range of a month to broad for your needs?

Nope. I'm pretty easy to please as long as we communicate dates and are able to plan on our side. (So the rest of this comment is just opinion that I don't mind being overridden.)

> ... EOL v0.10 less than a week after ... changes have landed.

If everybody has been expecting the EOL on a particular date then they should already have their lifeline in place: some newer version of Node.js. I also don't see a problem with a release _after_ Maintenance ends for conscience reasons, but the community libraries will have moved on.

If a CVE shows up in your inbox one day before drop of support do you release a new version? One hour? Does every new CVE trigger a push of the sunset by N days?

If we can't reasonably get a release out before the sunset date of that Node.js version my (harsh) opinion is "then never patch it." We promised support up until that date and no longer; the user should have been prepared to jump to the next version.

I also always assume that hacking groups sit on zero-days until the day after support ends for particular software versions. With that mindset patching something days or hours before dropping support simply means that it's safe until the day after we stop supporting it. :trollface:

mhdawson:
I'm thinking we can set the Maintenance to EOL as a fixed date as well.  If for some exceptional reason we believe we need to cut a release after that then I don't think that prevents us from doing that.

README.md
```diff
@@ -28,8 +28,8 @@
   <td><b>Active</b></td>
   <td>v4</td>
   <td>2015-10-01</td>
-  <td>2017-04-01</td>
-  <td>2018-04-01</td>
+  <td>April 2017</td>
```
vip2017A: Dadf

mhdawson:
@MylesBorins I think this is waiting on an update or response from you.

MylesBorins:
So it looks like this never got discussed in an LTS meeting. Lets discuss this in the next one and land it

gibfahn:
This was discussed in the [last LTS meeting](https://github.com/nodejs/LTS/issues/213) and there were no objections, so this is good to land.

README.md
```diff
@@ -61,6 +61,10 @@ period of 18 months from the date it enters LTS coverage. Following those 18
 months of active support, the major version will transition into "maintenance" 
 mode for 12 additional months.
 
+The exact date that a release stream will be moved to LTS, moved between LTS
+modes, or deprecated will will be chosen no later than the first day of the month
+it is to be changed with no less than 14 days notice.
```
gibfahn: `of the month it is` -> `of the month. It is`

MylesBorins:
rebased and updated based on new calendar. Can people please take a look

screen shot 2017-05-24 at 12 30 27 pm

README.md
```diff
@@ -6,13 +6,13 @@
 |   :--:  |    :---:      |   :---:  |       :---:      |       :---:       |      :---:      |
 |  v0.10  |**End-of-Life**|    -     |        -         |    2015-10-01     |   2016-10-31    |
 |  v0.12  |**End-of-Life**|    -     |        -         |    2016-04-01     |   2016-12-31    |
```
gibfahn: While you're here, is there a reason these are `v0.10` and `v0.12` instead of `0.10.x` and `0.12.x`?

Issue #107, opened 2016-05-16
make the LTS schedule more accessible


oe:
i've been thinking about this. even though the website home has a link to the LTS schedule graphic, it's still not _the_ most accessible document for figuring out release lines. it has a whole bunch of colors, bars, lines and indicators, and it takes a while to find what you're looking for.

because the most prominent use-case for the schedule is to find out _which version_ and _when that version_ enters LTS and Maintenance, there should be an application to find this information quickly that's also relative to the current date (not absolute, like the graphic).

to be fair, i made something that does this. it can be found at http://nver.tentakel.net/ (source at http://github.com/tentakel/nver). although it's just a prototype right now, i think it makes it a lot easier to understand the release cycles and find the information you're looking for.

so the question is: is this a good method of simplifying the schedule? or is it absolute rubbish and the LTS graphic is better?

TheAlphaNerd:
@tentakel have you seen https://github.com/nodejs/LTS/pull/104?

tentakel:
@TheAlphaNerd yep, but that's still:

- a whole wall of text to read
- a table/diagram with relative dates

although it's simplified (which is good!), it's still not really "easily" accessible

boneskull:
@ore I like it. 

I agree the updated graphic is trying to convey too much information at once. For example: arrows showing what changes go from this place to that, and the presence of "master" at all. I'm also inferring what the arrows mean, because there is no key. I have no idea what the dashed lines mean.

mansona:
Hey folks, I have 2c to weigh in here. I'm currently looking for "versions of node that are still *maintained*" i.e. have not EOL yet. 

Am I right in saying that the only "Active" (i.e. getting security patches) Node versions are v0.10, v0.12, v4.x and v6.x. I know v5.x has officially gone EOL. 

I guess i agree with @Lac on this one, the whole thing is not _clear_ at the moment. My use case: I want to test using Travis on *all supported Node versions* and I just need a list, a simple list of versions that I need to test on. I am not sure if I've interpreted the PNG correctly because of the "LTS Status" doesn't really match with "is currently supported". 

Does this make sense?

boneskull:
Indeed, the above would be helpful for me as well. Are *any* versions of iojs maintained?  I'm guessing no.

mansona:
@boneskull no there aren't any iojs maintained but if I can remember correctly Node 4.x is really not that far from what iojs was. It would be worth doing a 1-for-1 replacement and seeing if you have any issues.

[...]

PR #157, opened 2016-11-15
doc: release schedule as JSON

rvagg:
Idea floated in #107 

The future dates for `"start"` and `"lts"` and future `"codename"` (for "v8" at this stage but we'd push that out to include `"v9"` when it's in sight) will be filled in when we have them, they should be considered placeholders at the moment (i.e. the actual date we release v8 is unknown but we know it'll be in April, same as the switch-to-LTS point).

@nodejs/lts


fene:
this looks good! can we build an official website akin to http://github.com/fene/nver with it? i feel like it's fairly simple and makes sense

rvagg:
the README here still feels like the right place for that since we link to it from so many places, and it'd be cool to still promote third-party sites if they do a better job too

srl295:
Now i'll have to update [ICU's](http://icu-project.org/xml/icumeta.json)

rvagg:
live

leem23m (approved):

PR #162, opened 2016-11-28
update schedule image for 2016

rvagg:
A quick update of the image, instead of trying to do this as part of the full README rewrite.

shifts the timeline across by quite a bit, includes up to v8, drops v0.10, includes the additional maintenance period for Current

@nodejs/lts


Fishrock123:
@rvagg Can we switch the names to e.g. "Node 8"? or "Version 8"?

Also for a future idea maybe, but wouldn't it make more sense to have the newest branch closest to `master`?

jasnell:
+1 to dropping the `v` prefix in the versions in the chart. 
Since 0.10 is dropped, this should also drop Node.js 5.x given that it is EOL also.
Also, perhaps shift the start of the timeline over to start at April 2016?

hbetts:
+1 for dropping the `v` prefix.
+1 for dropping Node v5 given that it's reached EOL.
~ On whether to shift the start time reference to April 2016.

Can a visual indicator be added to the v8 maintenance block so that it's apparent to viewers that the maintenance window isn't shorter than the other Node LTS releases?

rvagg:
Hey folks, please check out the latest version of this. I've just dropped 0.10 and 0.12 and the 'v' on the version names. v5 is still there and the timeframe starts at the same place. I'll see if I can get someone cleverer here to rejigger it a bit more radically if I can't come up with something better myself. In the meantime, can we get this one merged just to get the graphic current?

egoroof:
What does actually "Master" show? It isn't a version and it will be forever the same. I think it's better to drop it too. What do you think?

sam-github:
LGTM, though if you feel like it, I think it makes more sense for master/unstable to be at the bottom, so from top to bottom it goes from oldest to newest (unstable is newest). Or drop it entirely as suggested above, I think unstable builds from master are pretty self-explanatory.

jasnell:
This LGTM. I am curious, however ... it might be possible to rig up a URL endpoint that serves this up more dynamically. Will have to stew on that.

rvagg:
OK, taking this as general +1 for this as an interim solution but we have things to work toward. Dropping master makes sense, it's a remnant from the original version that showed where commits come from which was important when we first formulated the flow of commits along with the schedule. Dynamic endpoint is a nice idea too but non-trivial. I could potentially bolt this on to nodei.co code which does this kind of thing already. SVG would be nice but .. cross-browser fonts and SVG make it basically unusable for anything containing text.

PR #165, opened 2016-12-08
doc: changes must be in Current for 2 weeks

### LINKED ISSUE #164: Document that changes must be in Current before backporting
As noted by @TheAlphaNerd [here](https://github.com/nodejs/node/pull/9139#issuecomment-264543910), changes must be in a _Current_ release for 2 weeks before they may be backported.

I could not find that documented on the LTS [README](https://github.com/nodejs/LTS).

hbetts:
Closes #164

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,10 @@ Changes in an LTS-covered major version are limited to:
 5. Certain performance improvements where the risk of breaking existing applications is minimal;
 6. Changes that introduce large amount of code churn where the risk of breaking existing applications is low and where the change in question may significantly ease the ability to backport future changes due to the reduction in diff noise.
 
+Generally changes are expected to live in a *Current* release for at least 2
+weeks before being backported. It is possible for a commit to land earlier at
+the discretion of the LTS Working Group and the maintainers of the LTS branches.
+
 Once a release moves into Maintenance mode, only ***critical*** bugs, 
 ***critical*** security fixes, and documentation updates will be permitted.
 


Fishrock123:
Unsure this will always play out to be true for certain things.

TheAlphaNerd:
this should likely have slightly more liberal language. some changes such
as doc or test changes do not need to live in current. other changes such
as experience breaking regressions can be escalated on a case by case basis

hbetts:
Extended the criteria under which changes may land, better defining those conditions under which changes may land without the 2 week bake time.

TheAlphaNerd:
I'm not a big fan of the language in this PR at the moment. To me it seems like it is a bit verbose, and is a bit too specific still.

I would much prefer something along the lines of

`Generally changes are expected to live in a Current release for at least two weeks before being backported. It is possible for something to land earlier at the discretion of the LTS Working group and the maintainers of the LTS branches`

Fishrock123:
Yeah that is better wording.

README.md
```diff
@@ -82,6 +82,12 @@ Changes in an LTS-covered major version are limited to:
 5. Certain performance improvements where the risk of breaking existing applications is minimal;
 6. Changes that introduce large amount of code churn where the risk of breaking existing applications is low and where the change in question may significantly ease the ability to backport future changes due to the reduction in diff noise.
 
+If a change updates documentation or tests, or resolves a breaking regression,
```
hbetts: @sam-github upon further thought, `breaking regression`, does not sound correct, and may not be meaningful. The only example of a non-breaking regression I could think of would be a regression introduced by a patch or minor change that removed documentation that had existed in an LTS release. That would be non-breaking in the sense that processes dependent on Node would continue operating, while there was a regression in the product shipped.

hbetts:
I've updated the pull request to be nearly verbatim what @TheAlphaNerd quoted (though I replaced `two` with `2` given that numbers (1,2,etc.) are used in the `README`.)

hbetts:
I usually lean towards more specificity, but the quoted text provided by @TheAlphaNerd is an improvement over the existing documentation, and, as also noted by @TheAlphaNerd, captures the particular subject of interest, which is the 2 week waiting period preferred by the LTS working group.

This change will hopefully help me, and others, with a better sense of what we should expect and when.

Thank you all for your help on this :-)

gibfahn:
@Fishrock123 @sam-github LGTY?

TheAlphaNerd:
landed in 1a10c515de

PR #178, opened 2017-01-09
make the LTS statuses in the table for Current clearer

Fishrock123:
I've always found "Not Available" a strange way to say "None" in this table.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@
   <td>2018-04-01</td>
 </tr>
 <tr>
-  <td>N/A</td>
+  <td>No LTS</td>
   <td>v5</td>
   <td colspan="3" align="center">N/A</td>
 </tr>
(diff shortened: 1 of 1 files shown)


richardlau:
I think in this case `N/A` meant "not applicable" since neither v5 nor v7 are LTS releases.

README.md
```diff
@@ -44,7 +44,7 @@
   <td>2019-04-18</td>
 </tr>
 <tr>
-  <td>N/A</td>
+  <td>None</td>
```
addaleax: tiny suggestion: I think `No LTS` might be even clearer :)

README.md
```diff
@@ -32,7 +32,7 @@
   <td>2018-04-01</td>
 </tr>
 <tr>
-  <td>N/A</td>
+  <td>End-of-Life</td>
```
richardlau: Since this is in a `LTS Status` column, `End-of-Life` may be misinterpreted as suggesting this was previously LTS (it never was).

Fishrock123:
Hmmm, looking again, maybe the v5 commit should be dropped?

PR #182, opened 2017-01-25
Release names

majgis:
I added a codename column to the LTS schedule table because it would be nice to have one authoritative list of codenames.  The addition of a column caused wrapping within the table, making it harder to read.  I solved that issue by shortening "End-of-Life" to "EOL".

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
   <th>Release</th>
   <th>Codename</th>
   <th>Active LTS Start</th>
-  <th>Active LTS End/Maintenance Start</th>
+  <th>Maintenance Start</th>
   <th>Maintenance End</th>
 </tr>
 <tr>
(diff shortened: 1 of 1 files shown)


hbetts:
Just my two cents, but could `Active LTS End/Maintenance Start` be shortened instead of changing `End-of-Life` to its abbreviation?

hbetts:
Personally, I like your v2 example. I find the table easier to visually parse given that the columns are more equally sized, and `End-of-Life` is spelled out (eliminating the cognitive overhead of translating that while reading).

majgis:
Fixed

Fishrock123:
Thanks, landed in https://github.com/nodejs/LTS/compare/46bca398ea7ed3b1ab88c65bc41063027172a597...0fafe811a73e4261d99f4b24dfc56a781fb8bbb2

PR #195, opened 2017-04-04
update version 8 LTS start year

cjihrig:


schedule.json
```diff
@@ -33,7 +33,7 @@
   },
   "v8": {
     "start": "2017-04-30",
-    "lts": "2016-10-31",
+    "lts": "2017-10-31",
```
richardlau: The readme currently says `2017-10-01`. These should be consistent.

hbetts: `schedule.ical` also refers to `2017-10-01`

gibfahn: @nodejs/lts should this be `10-01` or `10-31`? Either way let's fix and land.

MylesBorins: I think we should specify the month and no date. update day closer to

jasnell:
+1 to omitting the day. Just list the month and year

schedule.json
hbetts: +1 for emitting the day. Feels too specific, especially since LTS does not guarantee the LTS version will be cut on that day.

gibfahn:
@nodejs/lts any objections to just landing this as is? The question about day of the month is likely to be a moot point once/if https://github.com/nodejs/LTS/pull/141 lands.

PR #214, opened 2017-05-11
Updated year for lts in v8

mwelack:
I assume version v8 will be under lts starting 2017 not 2016 :)

sorry, just saw someone already made a pull request for this


gibfahn:
Good point, we should definitely land https://github.com/nodejs/LTS/pull/195

Issue #253, opened 2017-09-25
Date mismatches on meeting notes


mjomble:
Starting from 2017-04-24 (or possibly 2017-04-04), most of the meeting notes have different dates in the filename and inside the file.

gibfahn:
Good point, will fix.

Issue #235, opened 2017-07-17
LTS WG Meeting 17 July 2017


mhdawson:
I see by the schedule the meeting is set for today.  @gibfahn has been raising the issues lately but he is on vacation for a few weeks so I'm opening this one now (better late than never).

## Present

@nodejs/lts, maybe @nodejs/release if they're interested

## When

9PM Monday July 17th UTC, or in your local timezone:
https://www.timeanddate.com/worldclock/fixedtime.html?msg=Node.js+LTS+Working+Group+Meeting&iso=20170626T17&p1=179&ah=1

Previous meetings: https://github.com/nodejs/LTS/issues/233

## Where

Hangouts on Air:
- [Observers](https://www.youtube.com/watch?v=qFCsn63CV2A)
- [Participants] (will be added a few minutes before meeting starts)
- [Events page](https://www.youtube.com/my_live_events)
- [Minutes](https://docs.google.com/document/d/1FP6Lkb8zPbVBdWoGPF_wVee8L6TzlMYrxptoGJELe7A/edit?usp=sharing)

### nodejs/node

* CI: test with `--with-snapshot` on the slower platforms [#14222](https://github.com/nodejs/node/issues/14222)
* tracking issue: full VS2017 support [#13052](https://github.com/nodejs/node/issues/13052)
* (v6.x backport) src: allow CLI args in env with NODE_OPTIONS [#12677](https://github.com/nodejs/node/pull/12677)
* inspector: enable --inspect-brk in v6 [#12615](https://github.com/nodejs/node/pull/12615)
* Remove `indexOf` usage from tests in favor of `includes` [#12586](https://github.com/nodejs/node/issues/12586)

### nodejs/LTS

* doc: charter the Release working group [#223](https://github.com/nodejs/LTS/pull/223)

### nodejs/CTC

* Chartering the Release Team as a Working Group [#123](https://github.com/nodejs/CTC/issues/123)
* meta: charter the LTS Working Group [#122](https://github.com/nodejs/CTC/pull/122)

*Any questions in #node-dev on IRC or comments in this thread*

Please 👍 if you can make the meeting, and 👎 if you cannot.

MylesBorins:
Most of those issues are voting on semver minor changes

I'm going to suggest we skip this week if no one is opposed

sam-github:
I can make the time, but skipping seems reasonable, if we all comment on those issues in github.

mhdawson:
[...]

Issue #201, opened 2017-04-20
LTS Meeting April 24th - 5 pm EST


gibfahn:
Feel free to edit if I messed it up.

## Present

@nodejs/lts

## When

9PM Monday April 24th UTC, or in your local timezone:
https://www.timeanddate.com/worldclock/fixedtime.html?msg=Node.js+LTS+Working+Group+Meeting&iso=20170424T17&p1=179&ah=1

Previous meetings: https://github.com/nodejs/LTS/issues/194 https://github.com/nodejs/LTS/issues/185

## Where

Hangouts on Air:
- Observers: https://www.youtube.com/watch?v=CVPtof6dZLc
- Participants: https://hangouts.google.com/hangouts/_/ytl/9EDDsMjTm3qtKSlwGCeg8Pmasdbuo8UOT3DKbAxa5Zg=?eid=100598160817214911030&hl=en_US
- Events page: https://www.youtube.com/my_live_events

### nodejs/LTS
- Limit push access to LTS staging branches to backport team [#199](https://github.com/nodejs/LTS/issues/199)
- Node-v8 LTS support end on 2019-12-31 to meet EOSL of openssl-1.0.2 [#186](https://github.com/nodejs/LTS/issues/186)
- Backporting test fixes to maintenance branches [#205](https://github.com/nodejs/LTS/issues/205)
- Inspector - retain support for `--inspect --debug-brk` https://github.com/nodejs/node/issues/12364 (related: https://github.com/nodejs/node/pull/12615)
- Plan for triaging semver-minor backports in LTS meetings [#204](https://github.com/nodejs/LTS/issues/204)
- Potential Semver Minor Backports [#177](https://github.com/nodejs/LTS/issues/188)

### other

*Any questions in #node-dev on IRC or comments in this thread*

## Minutes

- doc: https://docs.google.com/document/d/1wyL9uQRJBSrUp4yD-bVIB51cRQQLOEbo9nRbz1O58nA/edit?usp=sharing

refack:
I'd like to join the meeting (I promise to to quiet).
Also as I said in #203, I it be an honor for me to join the LTS effort

gibfahn:
@refack FYI if you want to watch the meetings are all streamed live on Youtube. People do sometimes join "as observers" but I (personally) don't really know why that's done, WGs usually take questions (just post them in here).

hbetts:
[...]

PR #255, opened 2017-09-25
doc: fix meeting minute dates

gibfahn:
Fixes: https://github.com/nodejs/Release/issues/253

cc/  @mjomble, let me know if this LGTY. 

You can check the meeting issues here: https://github.com/nodejs/Release/issues?q=is%3Aissue+is%3Aclosed+label%3Alts-meeting

### DIFF
--- a/doc/meetings/2017-03-13.md
+++ b/doc/meetings/2017-03-13.md
@@ -3,7 +3,7 @@
 * Github issue: https://github.com/nodejs/LTS/issues/185
 * Meeting video: https://www.youtube.com/watch?v=w45MlYvMSQM
 * Previous meeting: https://docs.google.com/document/d/1YK6hYnaVAgGIYC50I8utpx7mD_L-E_127CuuymE3YvU/edit
-* Next meeting: April 4 2017
+* Next meeting: April 24 2017
 
 ## Present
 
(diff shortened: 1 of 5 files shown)


mjomble:
In the case of 2017-04-24.md, [2017-03-13.md](https://github.com/nodejs/Release/blob/master/doc/meetings/2017-03-13.md) suggests that April 4 might have been correct and it was the filename that was wrong.
However, I'm not sure, perhaps the meeting was actually later rescheduled for the 24th.

The rest looks good, assuming that these are the correct dates :)

mjomble:
Ah, from #201 it seems like 24 was indeed correct.

gibfahn:
>In the case of 2017-04-24.md, 2017-03-13.md suggests that April 4 might have been correct and it was the filename that was wrong.

Good point, adjusted the `Next meeting` dates too.

doc/meetings/2017-06-26.md
```diff
@@ -3,7 +3,7 @@
 - [Github Issue](https://github.com/nodejs/LTS/issues/233)
 - [Meeting Video](https://www.youtube.com/watch?v=xSo6YiKzc5M)
 - [Previous meeting](https://github.com/nodejs/LTS/issues/225)
-- Next meeting: 17 July 2017
+- Next meeting: 19 July 2017
```
mjomble: Should be 19 September?

gibfahn: Not sure what I was doing here, it should actually have been the 17th of July, see https://github.com/nodejs/Release/issues/235.

The meeting was cancelled, but we didn't know that at the time.

mjomble:
Looks good now :+1:

PR #266, opened 2017-10-26
meta: remove ical schedule

MylesBorins:
This cal event will not be able to be kept in sync with upstream
changes when added to someones calendar. It seems like the official
Node.js calendar would be a better place for this

/cc @williamkapke


williamkapke:
Ah- yes. This would be a great thing to track on the Node Foundation calendar.

For anyone that missed it, I recently documented calendar management access here:
https://github.com/nodejs/admin#nodejs-foundation-calendar

PR #277, opened 2017-11-01
update schedule image

cjihrig:


gibfahn:
I note that the new image is 960x500 rather than 2666x

cjihrig:
FWIW, the lts module also outputs SVG if you want something more scalable.

MylesBorins:
Would it be possible to include documentation as to how you updated the
image in this pr?

cjihrig:
@MylesBorins any suggestions on where those instructions would go? I don't really want to add it to the README, as I'd rather keep that document focused.

MylesBorins:
Could it be in the readme of the module, and just have a note below the image

cjihrig:
Sure. I added a link to the module's npm page here. That page already contains docs for how to use it, and the various supported options.

PR #287, opened 2017-11-21
doc: update table to show v0.10 EOL and v7 N/A

dado3310:
Update the table to reflect the current status of v0.10 and v7


MylesBorins:
This is a dupe of an old commit

PR #297, opened 2018-01-08
doc: current releases are still End-of-Life

gibfahn:
Now the repo covers Current releases as well, it should be clear about
their status.

Refs: https://twitter.com/wa7son/status/950300918347714560

cc/ @watson  @bnb

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,16 +2,16 @@
 
 ## Release schedule
 
-| Release | LTS Status          | Codename   | Active LTS Start | Maintenance LTS Start | Maintenance LTS End       |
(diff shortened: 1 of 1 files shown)


watson:
May I suggest modifying the date columns as well to show the dates for the uneven versions also? Here's an example of how it can be done: https://github.com/elastic/apm-agent-nodejs/issues/158

gibfahn:
>May I suggest modifying the date columns as well to show the dates for the uneven versions also? Here's an example of how it can be done: elastic/apm-agent-nodejs#158

That's a good idea, but I don't think I'm going to get round to it in the near future, so I'm spinning it into a separate issue: https://github.com/nodejs/Release/issues/303

Issue #303, opened 2018-01-22
Show dates for current versions


gibfahn:
>May I suggest modifying the date columns as well to show the dates for the uneven versions also? Here's an example of how it can be done: elastic/apm-agent-nodejs#158

See https://github.com/nodejs/Release/pull/297#issuecomment-355983919

PR #305, opened 2018-01-23
 doc: Add initial release date column to release table 

nephross:
Update release table with initial release date column and added end-of-life dates to non-LTS releases.

Fixes: https://github.com/nodejs/Release/issues/303


MylesBorins:
landed in 12cbb15b0d

Issue #317, opened 2018-03-29
Pick the next ten codenames or something like that


Trott:
As proposed by @ofrobots: It would be great if @nodejs/Release could select and publish the next 3, 4, 5, 10, 100 codenames so that we don't have a big discussion every time a destined-for-LTS release comes around. See https://github.com/nodejs/Release/issues/291 for a pretty good example of how out-of-hand this gets. Maybe it's just me, maybe this sort of discussion is fine, but I think it runs the risk of burying pretty much any other topic that needs to be talked about in a release.

Not sure if Release WG rules allow me to put this on the meeting agenda, but I'll label it as such and hope the answer is "yes". (There are Release-agenda and lts-agenda labels. Not sure if there's a difference.)

Trott:
Oh, and once picked, publish, of course.

Trott:
Here's my suggestion. Rubber-stamp it and I'll PR it into a document.

* Dubnium (10.x 2018)
* Erbium (12.x 2019)
* Fermium (14.x 2020)
* Gold (16.x 2021)
* Hydrogen (18.x 2022)
* Iron (20.x 2023)
* Krypton (22.x 2024)
* Lithium (24.x 2025)

By this point, we'll have hopefully decided to stop with the elements theme and let go of all the problems it entails: Many letters not represented, irksome to avoid poisons etc., often difficult or impossible to get a name that is familiar to the average English speaker ("Isn't Dubnium a music genre that is always at 114 bpm? Erbium? Was that named after Airbnb?")...

Maybe we can make a tradition of switching themes halfway through the alphabet every time.

ofrobots:
Those code-names sound good enough for me. A rose by any other name...

Specially if there is no conclusion on #127, or if conclusion is for the status-quo (continue using code names), I would much prefer that we decide the names upfront, instead of drowning out release planning conversation with bike shedding on the code name.

richardlau:
No objections here to those names.

olstenlarck:
> By this point, we'll have hopefully decided to stop with the elements theme and let go of all the problems it entails

It's cool that it isn't too late to switch theme now. I even think that it's better to switch it now than in 2025.

I don't know how this theme is chosen, but feels like it wasn't thought for these problems back then.

But anyway, i'm not against anything :) Just 2c.

cjihrig:
@Trott ship it :shipit:

mhdawson:
+1 from me on the proposed codenames.

[...]

Issue #291, opened 2017-11-28
Node.js 10.0.0 Tracking


jasnell:
FYI... I will be kicking off the process for Node.js 10.0.0 the first week of January. Current plan is to create the actual `v10.x` and `v10.x-staging` branches on March 1st.

mhdawson:
+1

jasnell:
Ping @nodejs/release @nodejs/lts @nodejs/tsc .@nodejs/collaborators.. Because Node.js 10.x will be an LTS branch, it needs an LTS code name. The 8.x code name was "Carbon", which means we're on the D's. Unfortunately, the options for `D` are quite limited.

* `Darmstadtium` - "the synthetic chemical element of atomic number 110. It is a superheavy metal that decays in thousandths of a second."
* `Dubnium` - "the chemical element of atomic number 105, a very unstable element made by high-energy atomic collisions."
* `Dysprosium` - "the chemical element of atomic number 66, a soft silvery-white metal of the lanthanide series."

Aside from the horrible names, it would be exceedingly funny to name a Long Term Support line after either an element that decays in thousandths of a second, or after a highly unstable element, so either of the first too automatically get my vote.

Please weigh in on your choice.

ofrobots:
My vote is none of the above. Let's pick another theme.

jasnell:
> My vote is none of the above. Let's pick another theme.

aww ;-)

kapouer:
??? Diamond !!!

trivikr:
[Deuterium](https://en.wikipedia.org/wiki/Deuterium)?
It has a number of commercial and scientific uses, like nuclear reactors, spectroscopy, tracing, drugs etc

MylesBorins:
The element system will eventually fall apart as there are not elements for
every letter.

It might be worth considering another naming scheme

jasnell:
Sure. Any suggestions?

bengl:
Is there actual value in still having a naming scheme? The numbers seem to be used basically everywhere.

MylesBorins:
Animals
Comic book characters
Cities
Spices
Movie titles

ronkorving:
- (Computer?) Scientists
- Stars and other celestial objects
- Star Wars characters (I think trademarks may be an issue though, same may be true for comic books)
- Food (Android is already exhausting confectionery names though)

Trott:
* Muppet characters (Animal, Beaker, Camilla, Dr. Teeth...)
[...]

PR #318, opened 2018-03-31
add CODENAMES.md

Trott:
Fixes: https://github.com/nodejs/Release/issues/317

### DIFF
--- a/CODENAMES.md
+++ b/CODENAMES.md
@@ -3,7 +3,7 @@ This file contains a list of upcoming codenames for LTS releases.
 * Dubnium (10.x 2018)
 * Erbium (12.x 2019)
 * Fermium (14.x 2020)
-* Gold (16.x 2021)
+* Gallium (16.x 2021)
 * Hydrogen (18.x 2022)
 * Iron (20.x 2023)
 * Krypton (22.x 2024)
--- a/CODENAMES.md
+++ b/CODENAMES.md
@@ -10,3 +10,5 @@ releases are subject to change.
 * Gallium (16.x 2021)
 * Hydrogen (18.x 2022)
 * Iron (20.x 2023)
+
+The release schedule is available as a [JSON](./schedule.json) file.
--- a/CODENAMES.md
+++ b/CODENAMES.md
@@ -1,5 +1,8 @@
-This file contains a list of upcoming codenames for LTS releases.
+This file contains a list of codenames for LTS releases.
 
+* Argon (4.x 2015)
+* Boron (6.x 2016)
+* Carbon (8.x 2017)
 * Dubnium (10.x 2018)
 * Erbium (12.x 2019)
 * Fermium (14.x 2020)
--- a/CODENAMES.md
+++ b/CODENAMES.md
@@ -9,5 +9,3 @@ This file contains a list of codenames for LTS releases.
 * Gallium (16.x 2021)
 * Hydrogen (18.x 2022)
 * Iron (20.x 2023)
-* Krypton (22.x 2024)
-* Lithium (24.x 2025)
--- a/CODENAMES.md
+++ b/CODENAMES.md
@@ -1,0 +1,10 @@
+This file contains a list of upcoming codenames for LTS releases.
+
+* Dubnium (10.x 2018)
+* Erbium (12.x 2019)
+* Fermium (14.x 2020)
+* Gold (16.x 2021)
+* Hydrogen (18.x 2022)
+* Iron (20.x 2023)
+* Krypton (22.x 2024)
+* Lithium (24.x 2025)
--- a/CODENAMES.md
+++ b/CODENAMES.md
@@ -1,4 +1,5 @@
-This file contains a list of codenames for LTS releases.
+This file contains a list of codenames for LTS releases. Codenames for future
+releases are subject to change.
 
 * Argon (4.x 2015)
 * Boron (6.x 2016)
--- a/README.md
+++ b/README.md
@@ -113,10 +113,8 @@ Note that while it is possible that critical security and bug fixes may lead to
 *semver-major* changes landing within an LTS stream, such situations will be
 rare and will land as *semver-minor* bumps in the LTS covered version.
 
-All LTS releases will be assigned a "codename" drawn from the names of elements
-on the Periodic Table of Elements. For each upcoming LTS release, the Release
-working group will select a handful of candidate names and submit those for a
-collaborator vote.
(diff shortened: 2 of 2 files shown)


olstenlarck:
Only left, writers of articles to use them ;d

kyrylkov:
@jasnell https://github.com/nodejs/Release/issues/317#issuecomment-377116324

jasnell:
Ok, so it's still an open discussion then. Rather than committing to these names for the next number of years, I'd rather we just come up with a better scheme now.

shelleyp:
As I noted elsewhere #291 lordie, what is the issue now?

Other than Alex's legitimate concern (gold does have unique connotations) the list is toasty. 

More importantly, you'll be done. Done. A not-that-consequential item will actually be resolved, and you can then move on to the what-will-we-ultimate-do-about-buffer-constructor issue.

Trott:
> Rather than committing to these names for the next number of years, I'd rather we just come up with a better scheme now.

Those two options are not mutually exclusive. Sure, let's come up with a better naming scheme. But in the meantime, let's have a list of names that don't need to be debated until that new scheme comes. Approve this, and you still don't have to debate names for individual releases even if conversation about whether the theme should be Futurama characters or Simpsons characters drags on for 7 years. 

Literally. You won't need a new theme until 2025. But if you manage to select a new theme before 10.x, then great. You can throw this out at that time.

Trott:
@shelleyp By the way, I updated Gold to Gaiilium as suggested by @rauschma in #317.

jasnell:
@Trott ... I'll look at this more tomorrow after the holiday. We're still 23 days out from the release so there's no reason to rush.

Trott:
> @Trott ... I'll look at this more tomorrow after the holiday. We're still 23 days out from the release so there's no reason to rush.

@jasnell Waiting a day is no big deal, and I'd rather get more @nodejs/release approvals anyway.  But please let us not give even the appearance of slow-walking this.

We all complain (correctly) that it is impossible to keep track of everything that goes on across the nodejs project issue trackers. This is an opportunity to short-circuit predictable endless low-value bike-shedding. Failure to do this means exacerbating the very problem we claim to be interested in solving. Actions speak louder than words.

styfle:
I may have missed this part of the discussion, but are you intentionally skipping J?

That might confuse users who don’t know the complete list of code names.

drewfish:
@styfle There are no chemical element names that start with `J`, not in English at least. (Or not yet... perhaps someone will discover/name something by 2023 😄)

styfle:
@drewfish Yes I am [aware of the issue](https://github.com/nodejs/Release/issues/304).

But perhaps this PR should end at `Iron` and then we can bikeshed `J` at a future date. And it is possible something will be discovered in that time.

CODENAMES.md
```diff
@@ -0,0 +1,13 @@
...
+* Argon (4.x 2015)
+* Boron (6.x 2016)
+* Carbon (8.x 2017)
+* Dubnium (10.x 2018)
+* Erbium (12.x 2019)
+* Fermium (14.x 2020)
+* Gallium (16.x 2021)
+* Hydrogen (18.x 2022)
+* Iron (20.x 2023)
+* Krypton (22.x 2024)
```
styfle: Remove `Krypton` and `Lithium`.

Trott:
@styfle I agree on not assuming we'll skip J so I took out the K and L names. Now the list will get us out "only" 5 years to 2023.

Trott:
Hmmm...unfortunately @nodejs/release does not in fact contain everyone listed in the README as a member. (In fact, it is less than half the people listed!) Despite what the README says to the contrary, it looks like the full membership (or at least something closer to it) is in @nodejs/LTS. 

So, uh: @nodejs/LTS PTAL.

Trott:
Hmmm...my attempt to @-mention @nodejs/lts failed. Preview tells me this one will succeed. So trying again...

@nodejs/lts PTAL

bnb:
I am not a member of the Releases WG, but +1 from me. Thanks for taking the needed initiative here, @Trott ❤️

CODENAMES.md
```diff
@@ -0,0 +1,11 @@
+This file contains a list of codenames for LTS releases.
+
+* Argon (4.x 2015)
+* Boron (6.x 2016)
+* Carbon (8.x 2017)
+* Dubnium (10.x 2018)
+* Erbium (12.x 2019)
```
jasnell: Please mark everything for 12.x and higher as "tentative".

jasnell: That is... tentative in so far as we may come up with a different naming scheme.

gibfahn:
+1 on this, let's just have a set of names we can default to, then if there's ever enough of an agreement to change the names then we can do that (be it next week or in 4 years).

FWIW I think the best text name for Node 10.x is `Ten`, but that's just me 😁 .

However I think if we're going to have this it should be machine readable, so we can use it for backporting.

gibfahn:
>@gibfahn nvm reads it from the index.tab on nodejs.org/dist, is that not what you mean by machine readable?

That's fine, but I'm talking about having a tool for backporting that works out that the first LTS version of Node 8 (8.6.0 or whatever) should be called Node Carbon for creating the release, so it can't rely on that page.

I'm fine with that format though.

srl295:
OK… but isn't this already a field in https://github.com/nodejs/Release/blob/master/schedule.json ?  And in any event, having something first in json seems better than just text.

gibfahn:
>Isn't the codename already a field in /schedule.json@master ? Could we just add values to that (marked tentative)?

SGTM

CODENAMES.md
Trott: "Codenames for future releases are subject to change." This way we don't have to update this document every release (because you know we won't).

Trott:
Someone land this! (I'm not a member of the working group so I'm not going to do it because that would be rude.)

CODENAMES.md
```diff
@@ -0,0 +1,12 @@
+This file contains a list of codenames for LTS releases. Codenames for future
+releases are subject to change.
```
srl295: It just hit me that this is the same *repo* as `./schedule.json`. Could we at least mention that file or even have a link from there?

srl295: also README.md here mentions the codenames and could link to CODENAMES.md

Trott:
@srl295 Added the links you requested. Everyone else might want to review. Also: Someone please land this!  😄

CODENAMES.md
yvettep321: Thank you

PR #321, opened 2018-04-04
doc: add process for adding new releasers

MylesBorins:
Formalizing process around adding releasers as we have some folks nominated


MylesBorins:
Hey @nodejs/release @nodejs/tsc  can people please take a look

We have two outstanding release nominations that I would like to onboard, next week at collab summit seems like a great time to do it.

PR #331, opened 2018-04-27
Add "Dubnium" Codename to v10.0.0 in schedule.json

caseyWebb:
If I understand correctly, this should make `lts/dubnium` work with NVM, and thus TravisCI and the many others that rely on it.

### DIFF
--- a/schedule.json
+++ b/schedule.json
@@ -48,6 +48,6 @@
     "lts": "2018-10-01",
     "maintenance": "2020-04-01",
     "end": "2021-04-01",
-    "codename": ""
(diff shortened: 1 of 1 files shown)


richardlau:
@nodejs/lts @nodejs/release Do we set the codename before v10 becomes LTS in October?

jasnell:
In the binary? No, the codename is not set until the actual LTS release. The codename applies only to that specific minor release and above

caseyWebb:
👍 didn't realize they have to graduate to LTS. Closing.

PR #344, opened 2018-06-08
doc: add missing step to process for new releasers

targos:


joyeecheung:
Oh oops, i thought this is the core repo

PR #345, opened 2018-06-10
doc: add Michaël Zasso to Releasers team

targos:
Ref: https://github.com/nodejs/node/pull/21114


targos:
Another option is to replace the duplicated list with a link to https://github.com/nodejs/node#release-team

Issue #361, opened 2018-10-01
Incorrect start date in schedule.json


bnb:
The `schedule.json` file asserts that Node.js v10 is LTS today. As far as I know (?), this isn’t the case.

Historically it seems we trend more toward the end of the month. If truly incorrect, would it be acceptable/more accurate to update it to the 31st and adjust down if needed? Happy to create a PR if the answer is yes.

For what it’s worth, I discovered this last night while working on [node-release-lines](https://GitHub.com/nw/node-release-lines), which uses the data from `schedule.json` rather heavily.

mhdawson:
@jasnell, @MylesBorins can you two comment if we have a specific date to update it to yet?

PR #363, opened 2018-10-01
Update v10 LTS start date

bnb:
Closes https://github.com/nodejs/Release/issues/361, as per @MylesBorins suggestion (https://github.com/nodejs/Release/issues/361#issuecomment-425997013).


iamstarkov:
Can this be merged?

iamstarkov:
> The exact date that a release stream will be moved to LTS, moved between LTS modes, or deprecated will be chosen no later than the first day of the month. If it is to be changed, it will be done with no less than 14 days notice.

I feel a bit odd about this rule in the README vs this PR still being opened

PR #383, opened 2018-10-30
doc: add link to Dubnium LTS release download via nodejs/nodejs-latest-linker#2

knksmith57:
note, nodejs/nodejs-latest-linker#2 needs to land before this link will work


knksmith57:
just a heads up that this is unblocked now 😄

Issue #376, opened 2018-10-24
Move 8.x into maintenance mode in January 2019


MylesBorins:
I was thinking. We are quite behind on backports for 8.x, and it has an end of life in December of 2019. Would we be better off moving it to maintenance mode in January 2019 and focusing our efforts on keeping 10.x up to date?

mhdawson:
I'd be ok with that option.  It seems like if we have to choose between catching up on 8.x and falling behind on 10.x or keeping up on 10.x, keeping up with 10.x is a better choice.

rvagg:
This is a shame, I'm not excited by this proposal, but given that I'm not currently actively contributing to backports for 8.x and we don't seem to be able to get more people to do it, it's hard to argue against this.

AliSawari:
how is it possible to have two active LTS releases at the same time? wouldn't that make some confusions for the new coming developers to choose? Also, what does Maintenance mode means? is that something like a Tag that indicates the final days of a release? what's the point of it? thanks in advance.

mhdawson:
@AliSawari we need to have overlap so people have time to transition from one LTS release to another.

Maintenance mode means that commits which have landed in master/Current releases are no longer actively backported to the version.  Instead only Critical and security fixes are backported.

AliSawari:
@mhdawson Thanks for answering! 
yeah I think you guys should keep up with updating 10.x
I think most of the user-base chooses the latest version, as 8.x been removed from the [Node.js Main Page](https://nodejs.org) and replaced by 10.13.0 LTS. so its better to end its life earlier this time :smile:

richardlau:
> 
> 
> This is a shame, I'm not excited by this proposal, but given that I'm not currently actively contributing to backports for 8.x and we don't seem to be able to get more people to do it, it's hard to argue against this.

I agree. I also wonder if we're not going to be in the same situation with 10.x and 12.x. 

I'm assuming we'll be somehow communicating the change of date ahead of 1 Jan 2019 (so it's not a surprise to anyone not following this repository)? (It's too late, but maybe it should have been with the 8.13.0 release blog).

boneskull:
Was/is there a larger conversation around backporting somewhere?  The impression I get is that it's often painful to do.  I have a feeling it's also rather thankless work.

From the angle of the [package maintenance](https://github.com/nodejs/package-maintenance) initiative, it'd be good to collect some knowledge around what works and what doesn't.

[...]

PR #391, opened 2018-11-19
schedule: move 8.x maintenance to January 1 2019

MylesBorins:
Closes: https://github.com/nodejs/Release/issues/376


MylesBorins:
I will land on Monday if there are no objections

kanongil:
The schedule png needs to be updated for this.

Issue #390, opened 2018-11-19
LTS release proposal for q4 2018 and q1 2019


MylesBorins:
## Q4 2018 Carbon (8.x) Release Schedule

Date | Release
------------ | -------
Nov 20th      | v8.13.0
Dec 4th        | v8.13.1-rc
Dec 18th      | v8.13.1

At the end of Q4 8.x moves to maintenance mode

## Q4 2018 Dubnium (10.x) Release Schedule

Date | Release
------------ | -------
Nov 27th      | v10.14.2-rc
Dec 11th       | v10.14.2
Dec 26th      | v10.14.3-rc

## Q1 2019 Dubnium (10.x) Release Schedule

Date | Release
------------ | -------
Jan 8th         | v10.14.3
Jan 22nd      | v10.14.4-rc
Feb 5th         | v10.14.4
Feb 19th       | v10.14.5-rc
Mar 5th        |  v10.14.5
Mar 19th      |  v10.15.0-rc

MylesBorins:
Does anyone have thoughts about this schedule. Would people be able to volunteer taking responsibility for these specific releases? That would include getting backports up to date and cutting the release / rc's.

Happy to mentor anyone interested in helping out

BethGriggs:
I'll volunteer for v8.13.1, happy to help out with 10.x in Q1 too - maybe we could rotate between releasers when it is just 10.x?

MylesBorins:
@targos do you want to include current releases here too?

targos:
@MylesBorins I think we can open another issue for current releases?

/cc @BridgeAR

rvagg:
Looks good to me. It seems like the aim is for roughly monthly releases but aligned to Tuesdays, I like that.

How about for releaser scheduling we just keep 1 release ahead, so we always know who's doing next. If @BethGriggs, you do the next v8, @codebytere I think might like to do the next v10, then we're 👌. After those releases let's make sure we slot someone in for the next v10.

It looks like @BethGriggs is now on the releaser list. @codebytere do you wanting to move to full releaser status or would you prefer to just do prep?

IMO it might be best for new releasers to do a Current or two before doing an LTS so we're not springing fresh GPG keys on users in LTS. @nodejs/release any thoughts on that as a policy? Does it matter?

MylesBorins:
@codebytere has already been nominated and approved for the release team 🎉

Since they have been working on backports for a couple months now I was
imagining they could work on preparing the release and I could sign /
[...]

PR #393, opened 2018-11-21
doc: update the releaser onboarding process and rules

rvagg:
Coming out of https://github.com/nodejs/Release/issues/390 I've done some updating of the process for adding new releasers.

Clean-up, plus I've Added:

* Guidance on SSH key complexity and how to make one
* Notes about SSH & GPG key compromise
* Rules for starting new LTS releases (start with a Current)

The one thing I'm not sure about, but is already in here, is why all new releasers get added to nodejs-private? Perhaps that should be a separate step so as to give us more flexibility in adding releasers outside of the TSC while not expanding nodejs-private access too much. Security releases are a major ordeal and more important than even LTS releases to get right. And in practice, very few of us even do them because of the massive workload and complexity involved—and past experimentation with multiple releasers coordinating hasn't gone well. Happy to leave this in here for now but it might be worth discussing.

/cc @nodejs/release, @nodejs/releasers & @nodejs/build

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -130,24 +130,66 @@ See "WG Membership" above.
 
 When considering adding a new releaser an email should be sent to the
 [Technical Steering Committee](https://github.com/nodejs/tsc) for approval.
-After approval the nominee will be assigned a mentor from the release team
+After approval the nominee will be assigned a mentor from the releasers team
 to help walk them through the process to learn how to prepare a release.
-The nominee will then be expected to prepare 1 release on any branch, which
-can be promoted by any other member of the release team. After this
-release the nominee will be considered a full member of the releasers
-team. The first release promoted by a new release team member must have
-a mentor from the current release team available to support during the process.
 
-At any point during this process any member of the Release WG
-can raise an objection to the TSC.
+The nominee will then be expected to prepare one release on any active release
+line, which can be tagged, signed and promoted by any other existing member of
+the releasers team. After this release the nominee will be considered a full
+member of the releasers team. The first release promoted by a new releasers
+team member must have a mentor from the current releasers team available to
+support during the process.
 
-When being officially added to the Releasers group the following must happen:
+At any point during this process any member of the Release WG can raise an
+objection to the TSC.
 
-* Added to Releasers team in the Node.js org (grants ci-release access)
-* Added to security-release team in the Node.js and nodejs-private orgs
-* SSH keys need to be added to the dist user on the www machine
-* GPG keys need to be added to the nodejs/node [README.md](https://github.com/nodejs/node/#release-team)
+After being being approved by the TSC, the new releaser must:
+
+* Be added to the GitHub [releasers team](https://github.com/orgs/nodejs/teams/releasers) in the Node.js org (grants ci-release access)
+* Be added to the GitHub [security-release team](https://github.com/orgs/nodejs/teams/security-release) in the Node.js and nodejs-private orgs
+* Have a single, high quality SSH key added to the "dist" user on the primary www server (see below for guidelines regarding SSH key quality)
+* Have a GPG key added to the nodejs/node [README.md](https://github.com/nodejs/node/#release-team)
 * Open a PR in [nodejs/docker-node](https://github.com/nodejs/docker-node/) to add gpg key to [node.keys](https://github.com/nodejs/docker-node/blob/master/keys/node.keys)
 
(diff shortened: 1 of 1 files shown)


GOVERNANCE.md
```diff
@@ -130,24 +130,76 @@ See "WG Membership" above.
...
+line, which can be tagged, signed and promoted by any other existing member of
+the releasers team. After this release the nominee will be considered a full
+member of the releasers team. The first release promoted by a new releasers
+team member must have a mentor from the current releasers team available to
+support during the process.
+
+At any point during this process any member of the Release WG can raise an
+objection to the TSC.
+
+After being being approved by the TSC, the new releaser must:
```
MylesBorins: Should this be after being approved or after being fully onboarded (finishing prepping first release?)

GOVERNANCE.md
```diff
@@ -130,24 +130,76 @@ See "WG Membership" above.
...
+
+* Be added to the GitHub
+  [releasers team](https://github.com/orgs/nodejs/teams/releasers) in the
+  Node.js org (grants ci-release access)
+* Be added to the GitHub
+  [security-release team](https://github.com/orgs/nodejs/teams/security-release)
+  in the Node.js and nodejs-private orgs
```
MylesBorins: Should we add a link to security private? https://github.com/orgs/nodejs-private/teams/security-release/

GOVERNANCE.md
```diff
@@ -130,24 +130,76 @@ See "WG Membership" above.
...
+* Have a single, high quality SSH key added to the "dist" user on the primary
```
MylesBorins: Is this something we should be storing in secrets or have more process around then simply copying to the dist server?

GOVERNANCE.md
```diff
@@ -130,24 +130,76 @@ See "WG Membership" above.
...
+
+For releasing on LTS release lines, releasers must:
+
+* Be added as members of the GitHub
+  [LTS team](https://github.com/orgs/nodejs/teams/lts)
+* Fully release at least one Current Node.js release; for practice, and to help
+  promote the new GPG key to the user ecosystem prior to it being required to
+  verify an LTS release
+
+### SSH key guidance
```
MylesBorins: This is great. I'd love a follow up with similar guidelines for gpg

GOVERNANCE.md
```diff
@@ -130,24 +130,76 @@ See "WG Membership" above.
...
+  [releasers team](https://github.com/orgs/nodejs/teams/releasers) in the
+  Node.js org (grants ci-release access)
+* Be added to the GitHub
+  [security-release team](https://github.com/orgs/nodejs/teams/security-release)
+  in the Node.js and nodejs-private orgs
+* Have a single, high quality SSH key added to the "dist" user on the primary
+  www server (see below for guidelines regarding SSH key quality)
+* Have a GPG key added to the nodejs/node
+  [README.md](https://github.com/nodejs/node/#release-team)
+* Raise an issue in
```
MylesBorins: This could alternatively be to send a PR to https://github.com/nodejs/docker-node/blob/master/keys/node.keys

GOVERNANCE.md
mhdawson: I'd say after finished prep of the first release

MylesBorins:
ping @rvagg would you mind if I updated this based on the comments?

rvagg:
yeah, sorry @MylesBorins, busy month. I'll get back to this soon but if you want to take over this you're more than welcome to.

MylesBorins:
I've rebased and updated. I kept the initial commit as identical to the original commit (modulo upstream changes)... made updates in a new comimt

PTAL

MylesBorins:
Worth mentioning this maintains the rule around having to promote a current release before LTS. We have had friction with this right now, but I also think that may be related to how many people we onboarded at once

GOVERNANCE.md
```diff
@@ -130,24 +130,66 @@ See "WG Membership" above.
...
+Ed25519 SSH key can be generated with the following command:
+
+```
+ssh-keygen -t ed25519 -a 100 -N ''
+```
+
+By default, the resulting private key will be placed in `~/.ssh/id_ed25519` and
+public key in `~/.ssh/id_ed25519.pub`.
+
+The public key an be shared with an existing member of the releasers team to be
```
targos:
```suggestion
The public key can be shared with an existing member of the releasers team to be
```

Issue #147, opened 2016-10-23
Clarify what is a "current LTS release"


mgol:
At the end of the README the "current LTS release" term appears:

> It should be stated that the abstraction layer (currently NAN) should support all current LTS releases. Given that Active LTS will overlap for a period of 6 months, this means that the abstraction layer will, at any given point in time, fully support a maximum of 2 LTS releases.

Nowhere else in the README it's stated what it means for an LTS release to be "current". From the context I guess it's about being (one of) the Active LTS but it shouldn't have to be guessed.

netsup99:
Hi,

I cleaned cache and installed again NPM install, but it did not help.

My temporary workaround to start  learn typescript in this case was to load solution from 
https://olepetterdahlmann.com/2016/08/19/angular-2-startup-project-in-visual-studio-code-and-visual-studio/
This solution(package.json) uses  typescript": "^1.8.10",

 I tried with same configuration to use typescript": "^2.0.3", It did not load bootstrap.

//...Mike

mgol:
@netsup99 It seems you posted that in a wrong issue.

mtheoryx:
@mgol Does it make more sense to use 'Active LTS' in this case?

> It should be stated that the abstraction layer (currently NAN) should support all currently Active LTS releases. Given that Active LTS will overlap for a period of 6 months, this means that the abstraction layer will, at any given point in time, fully support a maximum of 2 LTS releases.

mgol:
Sounds good to me.

-- 
Michał Gołębiowski

TheAlphaNerd:
I'm going to close this for now, please let me know if you have any other questions

mgol:
@TheAlphaNerd This was not just a question, I think the README is confusing as it is now and that "current LTS" mentions should be replaced with "Active LTS".

mgol:
(unless that was not the intended meaning but then it still should be corrected IMO)

mhdawson:
I think it could be clarified to:

```
all LTS releases that have not yet reached the end of the Maintenance mode.
```

mgol:
...which means I understood it incorrectly? :) Sounds good to me if that's the intention.

Fishrock123:
This seems like poor wording on our part, it should be "Active LTS Release", correct?

Issue #359, opened 2018-09-15
Defining a word for all current releases


bnb:
In discussing with a friend, I realized that we don't have an actual word for the corpus of releases that are currently being worked on / not EOL.

Here's a sentence as an example:

> Node.js versions 6, 8, 10 are the only [word] releases currently. Soon Node.js version 11 will be be too.

Both of us instinctively reached for "Active" but that's currently a word used as a part of LTS terminology.

> Node.js versions 6, 8, 10 are the only active releases currently. Soon Node.js version 11 will be be too.

As far as I know, we don't have a word that defines this. I think it's a useful word to define, as I've reached for some kind of word like this several times.

Some possibilities:
* Circulated
* Supported
* Operating
* Effective

vsemozhetbyt:
"Supported" seems to be the most exact and substantial word? Others seem a bit evaluative and fuzzy.

MylesBorins:
I always say maintained

targos:
I say supported.

BridgeAR:
Both, supported and maintained sounds good.

joyeecheung:
I'd go for supported instead of maintained to avoid confusion about "maintenance mode"

mhdawson:
I prefer supported.

DrEVILish:
Current Releases
Maintained Releases

bnb:
@joyeecheung valid point, agreed!

@DrEVILish extending @joyeecheung's point, we have the "Current" release line which is the release line that's in active development. Current releases (and maintained releases) could theoretically get easily confused.

Seems like `Supported` is the only word that doesn't collide with another term and has... support? 😁

targos:
@bnb what's the next step? Would you like to document this somewhere?

bnb:
@targos yes, documenting that somewhere would be ideal. It would be awesome if we could have a `Glossary` section of the `README.md` that included all the terms we use (including this) 🤔

rvagg:
I say "actively supported release lines" whenever I do security releases that need to cover them all. I think also if you're going to clarify this terminology you should be clear about "releases" and "release lines" too cause that gets messy.

BridgeAR:
I added this to the release agenda to decide on a specific wording that we then use consistently.

Trott:
"actively-supported release lines" seems good, although it raises the question what the "inactively-supported" or "passively-supported" release lines are. Does it mean "maintenance" release lines are covered or not?

[...]

PR #398, opened 2018-12-08
doc: remove confusing wording regarding NAN

targos:
Use "Active LTS" instead.

Fixes: https://github.com/nodejs/Release/issues/147

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ only landed in the `v4.x` branch when a new `v4.x` release is being prepared.**
 ### Node abstraction layer
 
 It should be stated that the abstraction layer (currently [`NAN`][]) should
-support all *current* LTS releases. Given that Active LTS will overlap
+support all Active LTS releases. Given that Active LTS will overlap
 for a period of 6 months, this means that the abstraction layer will, at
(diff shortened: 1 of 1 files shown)


richardlau:
We already use "active" as separate from "maintenance".

targos:
@richardlau yes, and that's what it is about here.

richardlau:
We would expect the abstraction layer to also support maintenance releases.

richardlau:
Based on https://github.com/nodejs/Release/issues/359 maybe use "supported" instead?

targos:
But the next sentence says a maximum of 2 LTS releases. If maintenance was included it would be 3.

richardlau:
Hmm. You're right but I'd certainly be very surprised if I was using maintenance LTS and `nan` dropped support for it before it went EOL.

MylesBorins:
I think that this is a mistake... I do not believe that nan plans to remove support for a version of node that is supported by node

/cc @nodejs/n-api

targos:
I thought about it... My conclusion is that we can remove this section. NAN is out of the scope of the Release WG.

MylesBorins:
I've updated the title to better reflect the content. Landing

PR #410, opened 2019-01-22
doc: update to reflect N-API

mhdawson:
Update to indicate that changes for N-API will be allowed
in Maintenance when necessary.

In the last release WG meeting I took the issue to open this.


mhdawson:
@vsemozhetbyt thanks for the suggestion, added it :)

PR #310, opened 2018-02-16
Moving myself to emeritus for LTS and CITGM members

jasnell:
I'm still on the releasers team, but I have not been active in either the LTS backporting efforts or CITGM and I will be focusing on more forward looking issues for the foreseeable future.

PR #412, opened 2019-01-29
reflect James's current status correctly

thefourtheye:
If I am not mistaken, either the Emeritus list has to be fixed
or the current members list.

cc @jasnell

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,6 @@ and CITGM team members listed below.
 - [@Fishrock123](https://github.com/Fishrock123) - Jeremiah Senkpiel
 - [@geek](https://github.com/geek) - Wyatt Preul
 - [@gibfahn](https://github.com/gibfahn) - Gibson Fahnestock
-- [@jasnell](https://github.com/jasnell) - James M Snell
 - [@mhdawson](https://github.com/mhdawson) - Michael Dawson
 - [@MylesBorins](https://github.com/MylesBorins) - Myles Borins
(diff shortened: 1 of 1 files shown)


jasnell:
I am still on the releasers team and should not be shown as emeritus there.

I am emeritus on citgm and LTS only.

thefourtheye:
> @thefourtheye was his emeritus status not for being a releaser? do you have a reference for this?

I just noticed an inconsistency in the doc (James was listed in both Emeritus and current members), I don't have any references, sorry.

---

> I am still on the releasers team and should not be shown as emeritus there.

> I am emeritus on citgm and LTS only.

Correct, this patch updates the doc accordingly.

richardlau:
In either case the non-emeritus lists in the readme are synced using `ncu-team` and reflect the current membership of the corresponding GitHub teams. 

James did remove himself from the lists in the readme under https://github.com/nodejs/Release/pull/310 but a subsequent sync using `ncu-team` added him back (https://github.com/nodejs/Release/commit/19985fe7c16c764805bed84ea4901488bec35cbf) because he is still a member of those teams. So probably the correct thing to do would be to update the teams and then resync using `ncu-team` otherwise he'll just be added back the next time we resync.

MylesBorins:
I've updated the appropriate teams. Thanks for the patch @thefourtheye, thanks for the receipts @richardlau, and thanks for all the hard work @jasnell

PR #420, opened 2019-03-14
Proposal to update staging branches

BridgeAR:
This is a proposal to update staging branches in a way that it should become easier to handle multiple release lines next to each other (semver-minor and patch releases). This is done by using two staging branches instead of one.

The minor one is the one in which each commit that should ever be backported should be targeted against. For patch releases we cut a release from the minor branch. That should a) reduce the conflicts for semver-minor releases and b) allow commits to land directly instead of having to wait until a release to be merged.

On top of that I loosened the rule that only LTS members should land PRs on the LTS branches. Having a few more people being able to do that seems like a good step forward. I personally would actually like to extend this by allowing any collaborator to land PRs against non-LTS staging branches.

I also changed the staging branches description not be generic. So far this only described the current state for the LTS staging branches even though the other release lines behaved AFAIK identical.

----

We discussed this in the today's meeting and this is the first documentation update to go in that direction. If we agree upon this, we'll have to also update the documentation in Node.js and update the staging branches accordingly.

@nodejs/releasers PTAL

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,21 +133,23 @@ even-numbered major release is cut. Depending on circumstances the project may
 decide to provide an update to the odd-numbered release after the cutoff. However, 
 there is no guarantee that any release will be made.
 
-### LTS Staging Branches
+### Staging Branches
 
-Every LTS major version has two branches in the GitHub repository: a release
-branch and a staging branch. The release branch is used to cut new releases.
-Only members of the @nodejs/releasers team should land commits onto release branches.
-The staging branch is used to land cherry-picked or backported commits from
+Every major version has three branches in the GitHub repository: a release branch
+and two staging branches. The release branch is used to cut new releases. Only
+members of the @nodejs/releasers team should land commits onto release branches.
+The staging branches are used to land cherry-picked or backported commits from
 master that need to be included in a future release. Only members of
-@nodejs/backporters should land commits onto staging branches.
+@nodejs/releasers should land commits onto staging branches.
 
-For example, for Node.js v4, there is a `v4.x` branch and a `v4.x-staging`
-branch. When commits land in master that must be cherry-picked for a future
-Node.js v4 release, those must be landed into the `v4.x-staging` branch. When
-commits are backported for a future Node.js v4 release, those must come in the
-form of pull requests opened against the `v4.x-staging` branch. **Commits are
-only landed in the `v4.x` branch when a new `v4.x` release is being prepared.**
+For example, for Node.js v10, there is a `v10.x` branch and a `v10.x-staging-patch`
+and `v10.x-staging-minor` branch. When commits land in master that must
+be cherry-picked for a future Node.js v10 release, all patch commits must be landed
+in the `v10.x-staging-patch` branch while all semver-minor and patch commits must be
+landed in the `v10.x-staging-minor` branch. When commits are backported for a future
+Node.js v10 release, those must come in the form of pull requests opened against the
+`v10.x-staging-minor` branch. **Commits are only landed in the `v10.x` branch when a
+new `v10.x` release is being prepared.**
 
 [Argon]: https://nodejs.org/download/release/latest-argon/
 [Boron]: https://nodejs.org/download/release/latest-boron/


README.md
```diff
@@ -133,21 +133,23 @@ even-numbered major release is cut. Depending on circumstances the project may
...
 master that need to be included in a future release. Only members of
-@nodejs/backporters should land commits onto staging branches.
-
-For example, for Node.js v4, there is a `v4.x` branch and a `v4.x-staging`
-branch. When commits land in master that must be cherry-picked for a future
-Node.js v4 release, those must be landed into the `v4.x-staging` branch. When
-commits are backported for a future Node.js v4 release, those must come in the
-form of pull requests opened against the `v4.x-staging` branch. **Commits are
-only landed in the `v4.x` branch when a new `v4.x` release is being prepared.**
+@nodejs/releasers should land commits onto staging branches.
```
richardlau: GitHub's UI is playing up for me at the moment, so I can't double check, but I don't think @nodejs/releasers is a superset of @nodejs/backporters, so this change would prevent those that are in @nodejs/backporters but are not in @nodejs/releasers from landing things on the staging branches.

README.md
```diff
@@ -133,21 +133,23 @@ even-numbered major release is cut. Depending on circumstances the project may
...
+
+For example, for Node.js v10, there is a `v10.x` branch and a `v10.x-staging-patch`
+and `v10.x-staging-minor` branch. When commits land in master that must
+be cherry-picked for a future Node.js v10 release, all patch commits must be landed
+in the `v10.x-staging-patch` branch while all semver-minor and patch commits must be
+landed in the `v10.x-staging-minor` branch. When commits are backported for a future
+Node.js v10 release, those must come in the form of pull requests opened against the
+`v10.x-staging-minor` branch. **Commits are only landed in the `v10.x` branch when a
+new `v10.x` release is being prepared.**
```
BethGriggs: Would we still have specific proposal branches/PRs for merging into v10.x?

targos:
We should make sure this approach works with our tooling or that we can adapt tooling to it. One main difference with the current process is that the staging branches would no longer just be `vx.y + commits for the next release`

jasnell:
Hmmm I'm not really convinced that managing three branches per release line is going to make things easier.

BridgeAR:
@targos 
> We should make sure this approach works with our tooling or that we can adapt tooling to it.

I don't think our tooling would be impacted by this at all currently. would you be so kind and elaborate in what way you fear it might interfere?

> One main difference with the current process is that the staging branches would no longer just be vx.y + commits for the next release

I am not sure I can follow. But it reminds me that I should probably clarify how the staging branches would be updated:

Let's say we currently have v11.0.0. So both staging branches start at this point. Now we pull in the minor and patch PRs in the minor staging branch and the patch PRs from the minor branch into the patch staging branch (this order is important to reduce unnecessary conflicts for the latter).
The next release turns out to be a minor one and we use the minor staging branch. As soon as the release is done, the patch staging branch is rebased to v11.1.0.
Then repeat landing the PRs as above but the release afterwards is a patch release.
So we create the release proposal for that one and create the release. The minor staging branch stays as it is.
We repeat landing PRs as above and the release afterwards is a minor one. Now we go back to the way we dealt with it for v11.1.0.

I hope that clarifies it? I can improve the documentation to make this clearer as well.

targos:
Thanks, it's indeed more clear to me. But then it seems like this should only apply to LTS release lines?
For Current, have always been deciding minor vs patch based on the commits present on staging, and I think it's fine. The process that you are suggesting requires to backport first to the minor staging branch so it would not make it easier to prepare Current releases.

BridgeAR:
@jasnell the `$version.x` branch does not require any change to today. There is a tiny overhead but the win is that we have significantly less conflicts. That is why I suggest this way of dealing with it. This way is especially powerful as soon as we automate our landing process further. But even today it should significantly reduce the work of us releasers depending on how the releases look like and what conflicts could happen.

README.md
BridgeAR: Yes, that part would not be impacted by my suggestion.

There is one more thing that we might want to add as well: commits should be backported from the next upper release line (if applicable). That way we'd again reduce conflicts and have less manual work. This is especially important for the LTS release lines.

BridgeAR:
@targos this indeed improves the situation more on the LTS branches than on current. It is only important for current if we'd have a patch release. It would be extra work otherwise.

So we could again improve the way a bit by only doing the extra work with landing the patch PRs on the patch branch if we might release a patch release as the next one (which could happen with security releases at any point of time but those would probably not cause a lot of conflicts). If we plan to release a minor as the following release, there is no point in updating the patch branch at all.

BridgeAR:
I am closing this since there was no consensus about the way we could improve our release process and we also do not look into following up on that with a release survey. It's therefore unlikely that something like this is going to become the standard any time soon.

Issue #405, opened 2019-01-03
[Tracking] Additional semver minor to 8.x during maintenance


MylesBorins:
There are a couple different features that we might want to ship before 8.x goes EOL

* [tsfn support for n-api](https://github.com/nodejs/node/pull/25002)
* various crypto features (/cc @tniessen)
* [namespaces](https://github.com/nodejs/node/pull/21551)

As such I think it does make sense for us to do one more Semver-Minor relase of the 8.x release line.

Thoughts?

mhdawson:
@MylesBorins I assume you mean 8.x in your original post as opposed to 10.x.

MylesBorins:
@mhdawson fixed, thanks

richardlau:
I'd like to get https://github.com/nodejs/node/pull/25447 in if we do another 8.x release. Let me know if it needs a manual backport.

MylesBorins:
@richardlau can you please backport?

richardlau:
> 
> 
> @richardlau can you please backport?

Done! https://github.com/nodejs/node/pull/25521

mhdawson:
@MylesBorins what timing are you thinking for the release.  Just want to make sure we sync with @gabrielschulhof to see if there is any chance it can also include the PR to make tsfn non-experiemental as wel as opposed that coming later.

MylesBorins:
Removing label as we had consensus to do this release.

We should keep this open for tracking

ljharb:
It would be *really* nice to backport the `instanceof Array` fixes from https://github.com/nodejs/node/pull/20250/files#diff-286202fdbdd74ede6f5f5334b6176b5cL320 into 8 (and even 6). It breaks jest, and cross-realm use cases, and I can't conceive of how the fix would cause any trouble.

MylesBorins:
@nodejs/collaborators quick ping. Is there anything specific you would like to see on 8.x?

BridgeAR:
I would like to include https://github.com/nodejs/node/pull/26599.

lpinca:
https://github.com/nodejs/node/pull/25938

Trott:
From https://github.com/nodejs/Release#release-plan:

> Once a major version enters LTS coverage, new features (semver-minor) may only be landed with consent of the Release working group. 

And...

> Changes in an LTS-covered major version are limited to:
> 
> 1. Bug fixes;
> 2. Security updates;
> 3. Non-semver-major npm updates;
> 4. Relevant documentation updates;
[...]

PR #426, opened 2019-03-20
doc: remove incorrect information from README

Trott:
The README had a list that was supposedly the only types of changes that
land in LTS branches. However, other types of changes have landed there
routinely. Rather than try to codify, removal is a tacit acceptance that
we land whatever Release WG deems appropriate.

Refs: https://github.com/nodejs/Release/issues/405#issuecomment-474517315
Refs: https://github.com/nodejs/Release/issues/405#issuecomment-474572632

@MylesBorins @devsnek

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ major version transitions to the Long Term Support plan.
 
 Every major version covered by the LTS plan will be actively maintained for a
 period of 18 months from the date it enters LTS coverage. Following those 18
-months of active support, the major version will transition into "maintenance"
+months of active support, the major version will transition into Maintenance
 mode for 12 additional months.
 
 The exact date that a release stream will be moved to LTS, moved between LTS
--- a/README.md
+++ b/README.md
@@ -94,19 +94,6 @@ Once a major version enters LTS coverage, new features (semver-minor) may only
 be landed with consent of the Release working group. No semver-major
 changes other than those required for critical security fixes may be landed.
 
-Changes in an LTS-covered major version are limited to:
-
-1. Bug fixes;
-2. Security updates;
-3. Non-semver-major npm updates;
-4. Relevant documentation updates;
-5. Certain performance improvements where the risk of breaking existing
-   applications is minimal;
-6. Changes that introduce large amount of code churn where the risk of breaking
-   existing applications is low and where the change in question may
-   significantly ease the ability to backport future changes due to the
-   reduction in diff noise.
-
 Generally changes are expected to live in a *Current* release for at least 2
 weeks before being backported. It is possible for a commit to land earlier at
 the discretion of the Release working group and the maintainers of the LTS branches.


README.md
```diff
@@ -94,29 +94,10 @@ Once a major version enters LTS coverage, new features (semver-minor) may only
...
-
 Generally changes are expected to live in a *Current* release for at least 2
 weeks before being backported. It is possible for a commit to land earlier at
 the discretion of the Release working group and the maintainers of the LTS branches.
 
-Once a release moves into Maintenance mode, only ***critical*** bugs,
-***critical*** security fixes, documentation updates, and updates to ensure
-consistency and usability of the N-API across LTS releases will be permitted.
-Unless a change is ***urgent*** it will be planned into a release once per
-quarter. Such releases will only be made when necessary.
```
richardlau: This paragraph needs to stay -- It basically defines what being in `Maintenance` mode is. We have been following this for Node.js 6.x.

Node.js 8.x is an oddity in that according to the original schedule it shouldn't be in Maintenance until April but it was [brought forward to January](https://github.com/nodejs/Release/issues/376) at the end of last year. So far nothing in https://github.com/nodejs/Release/issues/405 has actually landed in an 8.x release and the things being discussed there are not guaranteed to land either (in fact this very paragraph, unless changed, should be the guidance as to what is accepted).

Trott:
@richardlau I've restored that paragraph. PTAL.

Trott:
> I am not in favor of us just removing these constraints as a general guidance for maintenance LTS

That's kinda the problem, though. They are presented as constraints but treated as general guidance.

mhdawson:
How about just changing 

> Changes in an LTS-covered major version are limited to:

to 

> Changes in an LTS-covered major version are most limited to the following unless reviewed and approved by the Release Working Group members:

Trott:
> > Changes in an LTS-covered major version are limited to the following unless reviewed and approved by the Release Working Group members:

Doesn't Active LTS get all semver-patch and semver-minor changes if they cherry-pick cleanly?

MylesBorins:
Nope. Every commit goes through a separate review with different
expectation then current

Trott:
> Nope. Every commit goes through a separate review with different
expectation then current

Trott:
> Nope. Every commit goes through a separate review with different
expectation then current

I'm not sure what "separate review with different expectation then current" means, but I can also tell this is going nowhere so I'm going to close this.

Trott:
> but I can also tell this is going nowhere so I'm going to close this.

Thanks to anonymous friend-who-has-my-back, I'm back here to say: I've been tipped off that the text above sounds snippy and annoyed. And sure enough, it does. Sorry. Didn't mean it that way. Meant it more like this:

I was offering this as a quick remedy but it looks like this isn't going to fly so someone else will have to do a real fix on the doc. Oh, well. I'll close this.

Didn't want to be a jerk to the Release Team, whose heavy workload is inadequately appreciated!

PR #432, opened 2019-04-03
doc: sync team membership (remove rvagg, add jasnell)

rvagg:
removing myself from this team, it's been a good ride but I'm dead weight at this point and I'm happy to see there's energy in here for doing this work

`ncu-team` has added @jasnell during the edit, I presume that's a correct reflection of the team.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,6 @@ and CITGM team members listed below.
 - [@mhdawson](https://github.com/mhdawson) - Michael Dawson
 - [@MylesBorins](https://github.com/MylesBorins) - Myles Borins
 - [@othiym23](https://github.com/othiym23) - Forrest L Norvell
-- [@rvagg](https://github.com/rvagg) - Rod Vagg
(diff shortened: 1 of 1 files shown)


richardlau:
> `ncu-team` has added @jasnell during the edit, I presume that's a correct reflection of the team.

James said in https://github.com/nodejs/Release/pull/412#issuecomment-458404066 that he's emeritus in the LTS team.

jasnell:
I am active in the release team. Inactive in lts.

rvagg:
@jasnell the "release" team is just a parent of citgm, lts and releasers and has no members itself. This repo used to be "LTS" but its team was never renamed when the repo was, so we have a discrepancy now. You're a member of nodejs/lts, nodejs/backporters and nodejs/releasers. ncu-team has got the sync right, so maybe you want to revisit what teams you're a member of?

jasnell:
Of the three I should only be a member of nodejs/releasers

rvagg:
done, in the actual teams and here too

PR #471, opened 2019-08-20
Create .DS_Store

bryanworthington:


MylesBorins:
.DS_Store is a mac specific file we should not be included in the repo.

Thanks for opening the PR but I'm going to close this.

Issue #458, opened 2019-06-11
A proposal for changes to LTS


MylesBorins:
I think that we should switch active an maintenance and active time frames for 12.x and on

TL;DR

Current: 6 months
Active LTS: 12 months
Maintenance LTS: 18 months

this would significantly reduce the LTS burden. Thoughts?

AdamMajer:
From distribution point of view (SUSE), switching Active LTS with Maintenance LTS is not a problem. It's actually better from stability point of view. But it would always be a better for distros if Maintenance LTS would be increased. I do understand that this is not always possible due to 3rd party code (OpenSSL, for example)

So no objections here. :+1:

mhdawson:
I'm +1 for the switch provided we are flexible enough to get changes into Maintenance when it makes sense.

BridgeAR:
I suggest we'll discuss this in our next meeting.

sam-github:
+1

cjihrig:
Has this been settled on? Based on https://github.com/nodejs/Release/pull/477 landing, it seems like it has been, and this should be closed.

richardlau:
We need to update the general policy. Here's a PR: https://github.com/nodejs/Release/pull/492

Issue #439, opened 2019-04-29
Prevent duplicated "Current Release"


BridgeAR:
Both, v11.x and v12.x are both our "Current Release" right now and that seems confusing. I suggest we switch to `Current Maintenance` for v11.x from the date that v12.0.0 was released. That way it's clearer how we support the release line.

BethGriggs:
+1, then it will match what we set in the `schedule.json`. We should add a note somewhere clarifying what Current Maintenance means in terms of our commitments (whether it is exactly the same as LTS Maintenence or not)

mhdawson:
I agree we need to be careful with "Current Maintenance".  We've purposely kept the commitment of any maintenance for a Current after 6 months vague.  I think it may be most straight forward that we just have a single "current" and that current switch on the date of the new release.

BridgeAR:
Seems like this is nothing we ever found a real solution for. It's not ideal and it would be great to improve this but we did not follow-up on it for a long time. Thus, I'm closing this as likely won't fix.

PR #477, opened 2019-09-12
schedule: change LTS dates for v12.x

BethGriggs:
- Moves LTS start date of v12.x a day earlier to avoid having parallel current releases. This also allows time for the v13.x changelog to be finalised as Major changelogs are a branch-diff of `vN.x` and `vN-1.x`.

- Swaps the Active and Maintenance timeframes around for v12.x and v14.x - https://github.com/nodejs/Release/issues/458

- Avoid having two current releases (v13.x and v14.x)- #439

(I will regenerate the timeline/schedule once we've got consensus on the dates)


BethGriggs:
@richardlau I think the end dates should be 30th April for even-numbered releases - I've updated PTAL

okuryu:
@BethGriggs You might have forgot to update the schedule in README.

PR #482, opened 2019-10-07
readme: update schedule for v12.x, v14.x

BethGriggs:


README.md
```diff
@@ -6,9 +6,9 @@
 | :--:     | :---:               | :---:       | :---:          | :---:            | :---:                 | :---:                     |
 | [8.x][]  | **Maintenance LTS** | [Carbon][]  | 2017-05-30     | 2017-10-31       | 2019-01-01            | December 2019<sup>1</sup> |
 | [10.x][] | **Active LTS**      | [Dubnium][] | 2018-04-24     | 2018-10-30       | April 2020            | April 2021                |
-| [12.x][] | **Current Release** |             | 2019-04-23     | 2019-10-22       | April 2021            | April 2022                |
+| [12.x][] | **Current Release** |             | 2019-04-23     | 2019-10-21       | October 2021          | April 2022                |
```
addaleax: It [sounds](https://github.com/nodejs/node/pull/29694#issuecomment-539561886) like we may want to try to include V8 7.8 in v12.x, which is [supposed](https://www.chromium.org/developers/calendar) to become stable on the original date, i.e. Oct 22. Is there a particulr reason for moving this forward one day?

richardlau: It was to avoid having parallel current releases: https://github.com/nodejs/Release/pull/477

BethGriggs: I vaguely remembered an issue being raised by a tooling author requesting us to not have overlapping current releases - but I cannot find the link to that now. 

As the changelog for v13.x is a diff between v12.x and v13.x - if the v12.x release went out earlier on the same day the v13.x changelog would need to be updated on the day of the v13.x release (which I'd like to avoid if possible). I guess an alternative solution would be if we opened a v12.x-RC in advance of 22nd which we could base the v13.x changelog on?

PR #490, opened 2019-10-16
meta: document that LTS releases are semver-minor

jasnell:
The fact that moving to an LTS release is semver-minor was missed previously in the policy

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,34 @@ even-numbered major release is cut. Depending on circumstances the project may
 decide to provide an update to the odd-numbered release after the cutoff. However, 
 there is no guarantee that any release will be made.
 
+#### Marking a Release Line As LTS
+
+To mark a release line as LTS, the following changes must be made to `src/node_version.h`:
+
+* The `NODE_MINOR_VERSION` macro must be incremented by one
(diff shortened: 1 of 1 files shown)


jasnell:
I'm good with that also.

MylesBorins:
@richardlau I just noticed your comment after I landed this. Apologies for doing so without that comment being responded to. Would you like to revert or alternatively transition the specific content to the release guide? I think having the content here is worthwhile personally

MylesBorins:
@richardlau we can remove the explicit instruction in https://github.com/nodejs/Release/pull/496 and open a PR against core. Would you prefer that?

richardlau:
> 
> 
> @richardlau we can remove the explicit instruction in #496 and open a PR against core. Would you prefer that?

@MylesBorins That would be okay with me. 

I don't object to the instructions being in this repository but I do think they shouldn't be in the README as that's (also) read by users of Node.js trying to understand the Release policy (and not necessarily how that policy is implemented at a technical level). Core would seem to be the place for those instructions as we already created https://github.com/nodejs/node/blob/master/doc/releases.md and it would be preferable to have instructions for releasers all in one place.

PR #492, opened 2019-10-20
meta: swap active and maintenance periods

richardlau:
Refs: https://github.com/nodejs/Release/issues/458

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,9 +82,9 @@ When a new *odd-numbered* major release is cut, the previous *even-numbered*
 major version transitions to the Long Term Support plan.
 
 Every major version covered by the LTS plan will be actively maintained for a
-period of 18 months from the date it enters LTS coverage. Following those 18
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -82,9 +82,9 @@ When a new *odd-numbered* major release is cut, the previous *even-numbered*
 major version transitions to the Long Term Support plan.
 
 Every major version covered by the LTS plan will be actively maintained for a
-period of 18 months from the date it enters LTS coverage. Following those 18
+period of 12 months from the date it enters LTS coverage. Following those 12
 months of active support, the major version will transition into "maintenance"
-mode for 12 additional months.
+mode for 18 additional months.
```
cjihrig: It might be worth adding a note that prior to Node 12, those numbers were reversed. The note can be removed after Node 10 is EOL.

richardlau: Added.

Issue #495, opened 2019-10-21
"Current" should include v12.13


ljharb:
If Erbium was a semver-minor <sub>(I still maintain it should have been a 12.12.1 patch, but let's stipulate that it's a minor for the context of this issue)</sub>, then that means it added a feature.

"Current" both conceptually means "latest", and also on the website is described as "latest features" - 12.13 has a feature 12.12 does not, or else it wouldn't have been a semver-minor bump.

In other words, prior to v13.0.0 becoming Current tomorrow, I think that v12.13.0 should be considered Current.

This isn't urgent now - it only happens with every odd .0.0 release - but I think it's important that the release terminology and semver are consistent. When/if v14 goes into LTS, and v15.0.0 is not yet released, I'd expect/hope the latest v14 to be considered Current in the interim.

<sub>I also think "current" is a way better name than "stable", and although "latest" would be a better name still, I don't think renaming the release line is at all worth the churn it would cause</sub>

targos:
Technically, v12.13.0 **is** still the latest release. It is even linked at https://nodejs.org/dist/latest/
Did you open this issue because v12.12.0 is Current on https://nodejs.org/en/ ?

targos:
In the Node.js release schedule/policy, a release cannot be Current and LTS at the same time, but I agree this can be confusing for people who visit the website between the releases of v12.13.0 and v13.0.0.
I think the best way to fix this would be to remove the Current option from the site if `latestLts > latestCurrent`.

ljharb:
Yes, that's why I opened it (also because of ensuing discussion on twitter).

Where does it say in the policy that a release can't be Current and LTS at the same time?

MylesBorins:
I've opened a PR to fix the landing page for days like today

https://github.com/nodejs/nodejs.org/pull/2707

With this PR `Current` will not show if it is less than the latest LTS

@ljharb the process regarding branches transitioning is documented in https://github.com/nodejs/release#release-plan

Arguably the language is a bit off as it says `When a new odd-numbered major release is cut, the previous even-numbered major version transitions to the Long Term Support plan.`... implying the new major being cut triggers the transition... but that isn't the spirit of the program. Current and LTS are two very distinct phases for a branch and in turn define the release process and how we land new changes. All changes landing on an LTS release branch must live in a Current release for at least 2 weeks and as such it is impossible for an LTS release to also be current, it would upend our entire process.

Two actions we can likely do to rectify are 

[...]

Issue #430, opened 2019-03-27
What is the difference between Current and Active?


toymachiner62:
I'm curious what the difference between the Current and Active statuses of a Node.js version are.

My team is currently on version 9 and trying to determine whether to move to 10, 11, or 12 but i'm not sure what the statuses actually mean and there is nothing documented on the README about them (at least that I skimmed through and found).

BethGriggs:
Does this help? https://github.com/nodejs/node#release-types

MylesBorins:
Current is actively updated on a weekly or bi-weekly basis. It will be Semver-Minor or Semver-Patch depending on the backlog on master.

Active LTS has a monthly release cadence and will only contain changes that have already gone out in a Current release.

For production services I do not advise ever using Current. I would have advised being on 10 until 12 goes to Active LTS. As you are already on 11 you may want to upgrade to 12 soon after release, and hold off on future upgrades until 14 is in LTS.

cjihrig:
"Current" represents the current release line that receives all non-breaking changes that land in Node's master branch. The things that land in an active LTS or maintenance line are covered in https://github.com/nodejs/Release#release-plan.

> My team is currently on version 9

That version has been end of life for a while. Companies should generally stick to even numbered versions. I'd recommend moving to 10 now, and 12 once it goes LTS.

toymachiner62:
> Does this help? https://github.com/nodejs/node#release-types

Only partially. It doesn't say anything about `Active`

toymachiner62:
Closing as i've gotten my answer. Thanks all.

maksymkp:
https://nodejs.org/en/about/previous-releases#previous-releases

PR #517, opened 2019-12-24
doc: add 'Release Phases' documentation

BethGriggs:
Please let me know if you think this should be moved elsewhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ phases.
  the new feature supports migration to later release lines.
 
 Changes required for critical security and bug fixes may lead to *semver-major*
-changes landing within an LTS stream, such situations will be rare and will
+changes landing within a release stream, such situations will be rare and will
 land as *semver-minor*.
 
 The term 'supported release lines' will be used to refer to all release lines
--- a/README.md
+++ b/README.md
@@ -93,14 +93,14 @@ Working Group.
 
 ## Release Plan
 
-New semver-major releases of Node.js are branched from `master` every six
+New *semver-major* releases of Node.js are branched from `master` every six
 months. New even-numbered versions are released in April and odd-numbered
 versions in October.
 
 In coordination with a new *odd-numbered* major release, the previous
 *even-numbered* major version will transition to Long Term Support. The
-transition to Long Term Support should happen after the new major version is
-released in a semver-minor release.
+transition to Long Term Support will happen in a *semver-minor* release and can
+ happen either before or after the new major version is released.
 
 Every even (LTS) major version will be actively maintained for 12 months from
 the date it enters LTS coverage. Following those 12 months of active support,
@@ -110,8 +110,9 @@ months. See [Releases Phases](#release-phases) for details of which changes
 are expected to land during each release phase.
 
 The exact date that a release will be moved to LTS, moved between LTS modes,
-or deprecated will be chosen no later than the first day of the month it is to change.
-If the release team plans to change the release date, it will be done with no less than 14 days notice.
+or deprecated will be chosen no later than the first day of the month it is to
+change. If the release team plans to change the release date, it will be done
+with no less than 14 days notice.
 
 All LTS releases will be assigned a codename. A list of expected upcoming
 codenames is available in [CODENAMES.md](./CODENAMES.md).
--- a/README.md
+++ b/README.md
@@ -33,6 +33,9 @@ rare and will land as *semver-minor*.
  added at the discretion of the LTS team - typically only in cases where
  the new feature supports migration to later release lines.
 
+The term 'supported release lines' will be used to refer to all release lines
+that are not End-of-Life.
+
 ### End-of-Life Releases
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,6 +20,14 @@ Dates are subject to change.
 
 The Release schedule is available also as a [JSON][] file.
 
+### Release Phases
+
+There are three phases that a Node.js release can be in: 'Current', 'Active Long-term Support', and 'Maintenance'. Odd-numbered release lines are not promoted to LTS - they will not go through the 'Active LTS' or 'Maintenance' phases.
```
BridgeAR: Nit: it would be good to keep the 80 character limit as in the rest of the file.

MylesBorins:
Overall this LGTM but I am noticing a bit of a duplication of content between this new section and the [release plan section](https://github.com/nodejs/Release#release-plan).

Should we perhaps remove duplicated content?

BethGriggs:
@MylesBorins I'll try and morph the content into the release plan

BethGriggs:
I've updated this quite significantly to try and reduce duplication. The main changes are:

- Added release phases section. 
  - Hopefully would clarify  https://github.com/nodejs/Release/issues/430

- Moved and shortened the "Changes in an LTS-covered major version are limited to" list to the following text: 
> New features, bug fixes, and updates that have been audited by the LTS team and have been determined to be appropriate and stable for the release line. Changes required for critical security and bug fixes may lead to *semver-major* changes landing within an LTS stream, such situations will be rare and will land as *semver-minor*.

The above text is far less specific than what we had before, and defaults to "at the discretion of the release/LTS teams".  Please let me know your thoughts on removing the list of specifics changes we land such as npm updates, etc. 

- Added "supported" release lines term definition. - #359

README.md
```diff
@@ -15,6 +15,27 @@ Dates are subject to change.
 
 The Release schedule is available also as a [JSON][] file.
 
+### Release Phases
+
+There are three phases that a Node.js release can be in: 'Current', 'Active
+Long Term Support', and 'Maintenance'. Odd-numbered release lines are not
```
richardlau:
```suggestion
Long Term Support (LTS)', and 'Maintenance'. Odd-numbered release lines are not
```

BethGriggs:
Addressed recent comments from @richardlau, @MylesBorins - I think this is good to go now

README.md
```diff
@@ -15,6 +15,29 @@ Dates are subject to change.
 
 The Release schedule is available also as a [JSON][] file.
 
+### Release Phases
+
+There are three phases that a Node.js release can be in: 'Current', 'Active
+Long Term Support (LTS)', and 'Maintenance'. Odd-numbered release lines are not
+promoted to LTS - they will not go through the 'Active LTS' or 'Maintenance'
```
MylesBorins: I thought that odd number releases went through a short maintenance period? is that no longer true?

README.md
```diff
@@ -15,6 +15,29 @@ Dates are subject to change.
...
+ changes that land on `nodejs/node` master branch.
+ * Active LTS - New features, bug fixes, and updates that have been audited by
+ the LTS team and have been determined to be appropriate and stable for the
+ release line.
+ * Maintenance - Critical bug fixes and security updates. New features may be
+ added at the discretion of the LTS team - typically only in cases where
+ the new feature supports migration to later release lines.
+
+Changes required for critical security and bug fixes may lead to *semver-major*
+changes landing within an LTS stream, such situations will be rare and will
```
MylesBorins: this could happen in LTS or Current afaik
```suggestion
changes landing within a release stream, such situations will be rare and will
```

README.md
```diff
@@ -70,72 +93,30 @@ Working Group.
...
 
-Once a release moves into Maintenance mode, only ***critical*** bugs,
-***critical*** security fixes, documentation updates, and updates to ensure
-consistency and usability of the N-API across LTS releases will be permitted.
-Unless a change is ***urgent*** it will be planned into a release once per
-quarter. Such releases will only be made when necessary.
+In coordination with a new *odd-numbered* major release, the previous
+*even-numbered* major version will transition to Long Term Support. The
+transition to Long Term Support will happen in a *semver-minor* release and can
+ happen either before or after the new major version is released.
```
MylesBorins: We may want to open a follow up issue to track this, if I recall this is the exact thing that caused the "two current releases" thing.

As we will also be transitioning Active LTS releases to maintenance at the same time now we may want to have slightly more robust / clear policy here.

To be clear, not blocking... this should be a follow up.

README.md
BethGriggs: It is true according to the `schedule.json`- I think we've been purposely vague about it in the past. We can change that - maybe something like "release lines may receive critical fixes shortly after going EOL at the discretion of the Release team"? But I am not sure which is the best approach personally.

README.md
BethGriggs: Agreed - I think https://github.com/nodejs/Release/issues/495 had some of the conversation tied to it

PR #544, opened 2020-03-05
doc: update dates for v14.x

BethGriggs:


BethGriggs:
@sam-github, we should probably add v15.x to the table too, and that'd be a good time to update the graphic.  I'll leave that for a following PR.

PR #560, opened 2020-04-01
doc: delay 10.x maintenance, update schedule + SVG

MylesBorins:
Based on discussion in last Release WG meeting

```bash
$ npx lts -s 2019-9-30 -e 2021-10-31 -d "$PWD/schedule.json" -g "$PWD/schedule.svg"
$ npx svgo "$PWD/schedule.svg"
```
Screen Shot 2020-04-01 at 4 20 02 PM


richardlau:
Maybe move the start and end dates for the SVG? If we start sometime this year then End-of-Life Node.js 8 drops off the chart.

MylesBorins:
As the end of life was only a couple of months ago I was thinking it was
good to include. Good either way

MylesBorins:
I'm going to land this so that people are aware of the changes to the release schedule we agreed to last week.

I do genuinely believe that we should keep 8 on there but will open another PR for us to discuss shifting the schedule

Issue #561, opened 2020-04-01
Document process of updating schedule


MylesBorins:
We currently don't have this documented and I have to dig through past pr's to figure out how to update the SVG.

```bash
$ npx lts -s 2019-9-30 -e 2021-10-31 -d "$PWD/schedule.json" -g "$PWD/schedule.svg"
$ npx svgo "$PWD/schedule.svg"
```

I was thinking we should document this process somewhere or automate it with a script. Thoughts on how folks think we should do this

richardlau:
Maybe via Actions when the schedule is pushed to the master branch? (and possibly on a chron to shift the time window shown?)

PR #562, opened 2020-04-03
doc: update SVG

MylesBorins:
```bash
$ npx lts -s 2020-01-01 -e 2021-12-31 -d "$PWD/schedule.json" -g "$PWD/schedule.svg"
$ npx svgo "$PWD/schedule.svg"
```

Refs: https://github.com/nodejs/Release/pull/560

Screen Shot 2020-04-03 at 12 18 54 AM


MylesBorins:
This update drops 8.x from the SVG, which I'm not 100% is a good idea just yet. Not everyone is aware the 8.x is no longer maintained and I think it is good to keep it documented.

richardlau:
> 
> 
> This update drops 8.x from the SVG, which I'm not 100% is a good idea just yet. Not everyone is aware the 8.x is no longer maintained and I think it is good to keep it documented.

I'm fine either way. We should update the chart start and end points at some point (maybe when we do that should be part of the discussions in https://github.com/nodejs/Release/issues/561). Would be perfectly happen to wait until we release 14.x, for example.

dilyanpalauzov:
The SVG contains 1:1 the information in the table above the SVG in README.md.  That table says nothing about 8.

Issue #30, opened 2015-08-06
Status of v0.8.x?


chorrell:
Hi!

Apologies if this has already been covered elsewhere. I'm going on the assumption that support (e.g., security fixes) for v0.8.x will be dropped (unless that's happened already?) I can't find any explicit statement about this anywhere, but it seems to be the case and it's not included as an LTS release (which is good!).

I'm asking because I want to know if and when I can stop including a v0.8.x in the Docker node image (https://github.com/joyent/docker-node/issues/34). It should be noted that once dropped, users would still be able to pull a node v0.8.x image from the registry (e.g, `docker pull node:0.8.28` or `docker pull node:0.8`) there just won't be any 0.8 updates once I drop it from the GitHub repo.

bnoordhuis:
> I'm going on the assumption that support (e.g., security fixes) for v0.8.x will be dropped (unless that's happened already?)

I believe that's the case, the last v0.8 release was over a year ago.  v0.8 wasn't discussed in any of the LTS discussions either, as far as I'm aware.

Fishrock123:
Yeah, as far as I am aware 0.8 had already been put to rest once 0.12 came out.

chorrell:
OK cool. I'll make plans to eventually drop if from the docker-node image.

Thanks!

PR #624, opened 2020-10-26
Add v0.8 to the schedule

balupton:
Dates pulled from:

- https://github.com/nodejs/Release/issues/30
- https://nodejs.org/dist/v0.8.0/
- https://nodejs.org/dist/v0.10.0/

Reasoning is that I have an automated maintenance suite for my 100+ packages that has been updated to use this schedule to determine support for node.js versions (and will generate compiled code for each node version the package intends to support) - some utility packages go back to node v0.8 - as such rather than [manually adjusting](https://github.com/bevry/node-versions/blob/4908ac650a464ec4ef42adce121e07466a0fd9e6/source/index.ts#L69-L74) the results I receive from this, I would rather just add the details here.


schedule.json
```diff
@@ -1,4 +1,8 @@
 {
+  "v0.8": {
+    "start": "2012-06-25",
+    "end": "2013-03-11"
```
richardlau, line 4: https://nodejs.org/dist/v0.8.28/ was the last v0.8.x release in 2014: https://nodejs.org/en/blog/release/v0.8.28/

richardlau (approved):

BethGriggs (approved):

mhdawson (approved):
LGTM

PR #643, opened 2021-02-02
doc: fix headers in team groupings

danielleadams:
The headers on the team groupings appear that the LTS team houses backport, releasers, and citgm; so the header tags have been adjusted to match the actual groupings (lts, releasers, citgm).

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ and CITGM team members listed below.
 
 <!-- ncu-team-sync end -->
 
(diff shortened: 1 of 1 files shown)


richardlau (approved):

nschonni:
Does Backporters stay as an H3 because it's a sub-team of LTS?

danielleadams:
@nschonni yep! that's correct.

PR #672, opened 2021-05-21
Schedule update guidance

BethGriggs:
I figured we needed a home for documentation/guidance such as #561. For now, I think it makes sense for this file to only contain guidance related to this repository, while the specific steps of creating releases remain over in the `releases.md` file in Node.js core.

Open to alternative suggestions (or names) of where this documentation should live. I do think we need this written down somewhere, though.

### DIFF
--- a/doc/release-guide.md
+++ b/doc/release-guide.md
@@ -11,9 +11,9 @@ When making amendments to the release schedule the following files in this repos
  * `README.md` - The [Release Schedule Table](https://github.com/nodejs/Release#release-schedule) should be updated.
  * `schedule.svg` - The timeline graphic should be regenerated. See [Updating the schedule.svg](#updating-the-schedulesvg)
 
- # Updating the `schedule.svg`
+# Updating the `schedule.svg`
 
- The following commands can be used to regenerate the `schedule.svg`:
+The following commands can be used to regenerate the `schedule.svg`:
 
 ```sh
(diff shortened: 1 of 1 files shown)


richardlau (approved):
I reckon we could probably semi-automate this but let's get this written down first in a way that we don't need to search previous pull requests/commits for the command lines to run.

targos (approved):

doc/release-guide.md
```diff
@@ -0,0 +1,23 @@
...
+# Updating the Release Schedule
+
+When making amendments to the release schedule the following files in this repository should be updated:
+ * `schedule.json` - The JSON file should be updated to reflect the added or amended dates/entries.
+ * `README.md` - The [Release Schedule Table](https://github.com/nodejs/Release#release-schedule) should be updated.
+ * `schedule.svg` - The timeline graphic should be regenerated. See [Updating the schedule.svg](#updating-the-schedulesvg)
+
+ # Updating the `schedule.svg`
+
+ The following commands can be used to regenerate the `schedule.svg`:
```
bl-ue, line 16: Indented by 1 space, probably because of the list above.
```suggestion
# Updating the `schedule.svg`

The following commands can be used to regenerate the `schedule.svg`:
```

bl-ue (changes requested):

mhdawson (approved):
LGTM with changes suggested by other reviewers

bl-ue (approved):

PR #706, opened 2021-10-21
doc: update release status for Node.js 14 and 17

richardlau:
Node.js 17 has now been released and Node.js 14 is now in maintenance.


richardlau:
I'm planning to open a follow up for the 16 LTS transition.

PR #707, opened 2021-10-21
doc: update status for Node.js 16

richardlau:
Node.js 16 has now transitioned to Long Term Support with codename
'Gallium'.

---

Not to be merged until after the LTS transition (2021-10-26).


hsdgd (commented):
Hi

PR #723, opened 2022-02-09
 schedule: update graphic, add dates for Node.js 19 and 20

BethGriggs:


cjihrig (approved):

richardlau:
Was the graphic generated before the schedule was updated? Node.js 19 and 20 are not showing on it.

BethGriggs:
> Was the graphic generated before the schedule was updated? Node.js 19 and 20 are not showing on it.

Aware - that was intentional. Adding those to the image before the removal of Node.js 12 I think makes the timeline harder to read/follow.  Showing the 7 release lines at once just doesn't seem ideal:

schedule

BethGriggs:
To add, I tried modifying the dates earlier to cut off Node.js 20 - but any shorter windows it ends up rendering awkwardly:

image

richardlau (approved):

Nixinova:
> I think makes the timeline harder to read/follow. Showing the 7 release lines at once just doesn't seem ideal:

That graph looks fine enough in my opinion.

PR #729, opened 2022-02-23
Move myself from CITGM to CITGM Emeritus

bengl:
While I had an initial flurry of activity on the CITGM project (5 years ago I think?) I haven't been involved with it since then. While I don't want to rule out any future activity, it's probably well overdue that I step down from that team more officially since my activity there is zero and has been for years. It's certainly possible (maybe even likely?) that this should be reversed in the future, but I'd rather reflect reality now, even if it's several years late.

I couldn't find an offboarding process for CITGM specifically, so I assume this PR suffices. Once it lands, I'll remove myself from the @nodejs/citgm .

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,7 +221,6 @@ and CITGM team members listed below.
 <!-- ncu-team-sync.team(nodejs/citgm) -->
 
 - [@al-k21](https://github.com/al-k21) - Oleksandr Kushchak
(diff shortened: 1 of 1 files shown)


bengl:
Removed myself from the team.

luisglz190289:
bengl-citgm-emérito

octavio12345300 (approved):

PR #753, opened 2022-06-14
Rename 18.x LTS to Hafnium

textbook:
- Continues the `-ium` pattern in use since 10.x
- Hafnium has an atomic number of 72, which is 4 * 18, and its most abundant isotope has an atomic mass of 180
- Hafnium-based compounds have been used in semiconductors

### DIFF
--- a/CODENAMES.md
+++ b/CODENAMES.md
@@ -8,7 +8,7 @@ releases are subject to change.
 * Erbium (12.x 2019)
 * Fermium (14.x 2020)
 * Gallium (16.x 2021)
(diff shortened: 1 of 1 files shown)


Breakerzoe1 (approved):
CODENAMES.md

danielleadams:
Hi @textbook, thanks for the contribution! We have already named v18.x Hafnium throughout the project, so we won't be able to change it at this point.

We do, however, have an ongoing conversation regarding letter that don't have elements on the Periodic Table if you are interested in contributing: https://github.com/nodejs/Release/issues/304

PR #775, opened 2022-08-22
Update master branch references to main

alexmuller:
The default branch for all relevant repos in the nodejs organisation is now 'main'.

PR #776, opened 2022-08-24
fix(schedule): rename Master to Main

Trott:
Added a blank at the end for good measure.


Trott:
This is unnecessary if https://github.com/nodejs/Release/pull/775 lands, but in case that gets held up on other changes, this can land.

richardlau:
I've merged https://github.com/nodejs/Release/pull/775. As I mentioned there, the SVG file is generated by running https://github.com/nodejs/lts-schedule so the same applies here re. the newline at the end of the file -- if that is required then a change should be upstreamed otherwise it's going to get overwritten the next time the image is regenerated.

Issue #499, opened 2019-10-31
Trim active releasers list


rvagg:
There's a _lot_ of SSH keys in place allowing for release promotion, 40 all up. 28 of them are unlabelled and a lot of the ones with labels are for people I know don't do releases anymore. This functionality punches a pretty significant hole in our security perimeter that protects what we publish so I'd like us to get it locked down.

Can I ask this WG to clarify who can currently perform releases. Is https://github.com/nodejs/Release#releasers-team accurate or should even it be refreshed? It shouldn't be hard to check who has performed a [release](https://nodejs.org/en/blog/release/) in the past X months if that's a good way to do a refresh.

Secondly, can I get fresh SSH keys for each of these individuals? Your GitHub .keys is fine if it just contains one key, otherwise if you can specify which one, just one per person (it'd be awesome if it was a dedicated key but that's not strictly necessary). I'll get them in, labelled and dated, replacing everything that's there now.

MylesBorins:
Please keep me on there. Appropriate key included below in details section

<details>
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDCYSbr8uTzB4kbUvwesreEmPfSNQUbu/Ud5ltHJCnE6RHf9h6UZ4fRzDnIOk6PKy2C2jWqD9k/BIItWhWO7HXLGMwEdC29Bq7kw7+fFOMFEIEBndHdMDKsl+OWY/mwlHJ8oMeroh8/pk9cChWcXVPQhAyYrWVkaUeomLSHSYT7aZdXGOSpyWAPCd5RcSdfgdFhAZ0wpvfpFa//UV6ypxEvftXROqy9qYK+hmMdFWeBKpiTEpARxIHY0dcVX7SDnVFtiQDRJ2AF1BOQ7W2OEPf+5aGVAZNSIh18q3lWL83skSNsSWGNUdYhXiip/IO48JX1lZyPaUxp6vrZnK3nx+j3 mborins@mborins-macbookpro.roam.corp.google.com
</details>

Fishrock123:
I should probably be removed as a releaser.

<details>
<summary>I _think_ this was the SSH key that was used, but honestly have no idea</summary>
[...]

PR #784, opened 2022-09-27
chore: move inactive releasers to emeritus

Trott:
"Releasers have access to critical infrastructure in the project - this elevated access must be restricted to active releasers. Members of the releasers team should be offboarded when they no longer intend to prepare releases. As a guideline, offboarding should be considered if a releaser has not prepared a release in the past 12 months."

Refs: https://github.com/nodejs/Release/blob/main/GOVERNANCE.md#offboarding-releasers

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,13 +209,8 @@ and CITGM team members listed below.
 
 * [@bengl](https://github.com/bengl) - Bryan English
 * [@BethGriggs](https://github.com/BethGriggs) - Bethany Nicolle Griggs
-* [@BridgeAR](https://github.com/BridgeAR) - Ruben Bridgewater
-* [@cjihrig](https://github.com/cjihrig) - Colin Ihrig
-* [@codebytere](https://github.com/codebytere) - Shelley Vohr
 * [@danielleadams](https://github.com/danielleadams) - Danielle Adams
-* [@jasnell](https://github.com/jasnell) - James M Snell
 * [@juanarbol](https://github.com/juanarbol) - Juan José
-* [@MylesBorins](https://github.com/MylesBorins) - Myles Borins
 * [@RafaelGSS](https://github.com/RafaelGSS) - Rafael Gonzaga
 * [@richardlau](https://github.com/richardlau) - Richard Lau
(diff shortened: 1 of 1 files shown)


Trott:
@nodejs/releasers

README.md
```diff
@@ -231,7 +227,6 @@ and CITGM team members listed below.
 * [@BridgeAR](https://github.com/BridgeAR) - Ruben Bridgewater
 * [@bzoz](https://github.com/bzoz) - Bartosz Sosnowski
 * [@gdams](https://github.com/gdams) - George Adams
-* [@MylesBorins](https://github.com/MylesBorins) - Myles Borins
```
cjihrig, line 234: Was this supposed to be the entry from line 214 in the diff?

BethGriggs:
After this lands, it might be a good time to try and tackle https://github.com/nodejs/Release/issues/499 (which includes cleaning up lingering SSH keys from the dist server). We're probably well overdue on some housekeeping on there.

README.md
Trott: Yes, whoops! I'll fix it.

danielleadams (approved):

jasnell (approved):

MylesBorins:
I'd like to remain on the release team if possible. I am heading on paternity leave and will be out until January, so this can wait until I return. While I don't actively prepare releases I think it is prudent to keep an npm adjacent team member on the team in case of npm related releases being created. So alternatively someone else from the team could step up and get involved in my absence.

How do folks feel about that?

Trott:
> I'd like to remain on the release team if possible. I am heading on paternity leave and will be out until January, so this can wait until I return. While I don't actively prepare releases I think it is prudent to keep an npm adjacent team member on the team in case of npm related releases being created. So alternatively someone else from the team could step up and get involved in my absence.
> 
> How do folks feel about that?

The 12-months thing is presented as a "guideline" and not a hard-and-fast rule, so the Release team can certainly choose to make an exception like that.

BethGriggs (approved):

PR #787, opened 2022-10-06
chore: update onboarding/offboarding to include private Slack channel

Trott:


GOVERNANCE.md
```diff
@@ -154,6 +154,7 @@ After the nominee's first prepared release has been promoted the new releaser mu
 * Open a PR in [nodejs/docker-node](https://github.com/nodejs/docker-node/) to add their GPG key to [node.keys](https://github.com/nodejs/docker-node/blob/main/keys/node.keys).
 * Open a PR in [nodejs/release-keys](https://github.com/nodejs/release-keys) to add their GPG key via `./cli.sh add $KEY_ID`.
 * Open a PR in [nodejs/Release](https://github.com/nodejs/Release) to add themself under the "Releasers team" heading via `ncu-team sync README.md`.
+* Add them to the `#nodejs-release-private` team on the OpenJS Slack.
```
richardlau, line 157: I'm not in front my computer with Slack configured right now, but earlier when the private channel was created it was `#node-release-private`. Has it been renamed? FTR I'd prefer it to be as it is here (with the `#nodejs-` prefix) for consistency with the other Node.js channels on the OpenJS Slack.

Trott: I'm not in the channel, so I can't see it. But I am the one who created it and I definitely intended it to be `nodejs` and not `node` for the reason you state (consistency with existing channels). So if I messed that up, let's rename it. (And if people in there can't rename it, re-add me temporarily and I'll rename it and then leave again.)

richardlau (approved):

GOVERNANCE.md
targos: Is the plan to invite people after this lands (I'm not in the channel yet)

Trott: I thought I invited everyone who is a releaser into the channel but maybe I messed that up to! Anyone in the channel should be able to both audit existing users and add anyone missing. (I think they should be able to boot anyone who doesn't belong too, but if not, we can do the "add Rich momentarily again so he can do it".)

RafaelGSS (approved):
LGTM.

GOVERNANCE.md
RafaelGSS: Just added you, @targos.

targos (approved):

BethGriggs (approved):

PR #816, opened 2023-01-12
doc: add minutes for meeting 2023-01-12

ruyadorno:
Fixes: https://github.com/nodejs/Release/issues/814

### DIFF
--- a/doc/meetings/2023-01-12.md
+++ b/doc/meetings/2023-01-12.md
@@ -1,0 +1,59 @@
+# Node.js  Release WorkGroup Meeting 2023-01-12
+
+## Links
+
+* **Recording**: https://youtu.be/NCApl6VSWcU
+* **GitHub Issue**: https://github.com/nodejs/Release/issues/814
+* **Minutes Google Doc**: https://docs.google.com/document/d/1LC7GK2W9HGzLFY5UatSpm2f2pY5k83jfOhKbVsDmRyw
+
+## Present
+
+* Release team: @nodejs/release
+* LTS team: @nodejs/lts
+* Rafael Gonzaga: (@RafaelGSS)
+* Richard Lau (@richardlau)
+* Beth Griggs (@BethGriggs)
+* Ruy Adorno (@ruyadorno)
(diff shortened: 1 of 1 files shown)


doc/meetings/2023-01-12.md
```diff
@@ -0,0 +1,58 @@
...
+  * Let’s open a backport PR to explicit our intention of landing it in the LTS line
+  * Make sure there’s no potential impact to CI in the issue brought up by Jordan in https://github.com/nodejs/node/pull/45693
+* Major release rotation? [#765](https://github.com/nodejs/Release/issues/765)
+  * We need to update the docs to mention that we should make a call for releasers 3 months in advance.
+  * We can also automate that with a bot/Github Action if someone has the time to look into that.
+  * Rafael added automation to the Release Team slack channel
+* Release plan - v19.x Current [#793](https://github.com/nodejs/Release/issues/793)
+  * Michael signed up for a mid-February release
+* Release plan - v18.x Active LTS [#737](https://github.com/nodejs/Release/issues/737)
+  * Let’s reach out to Juan and make sure he’s still able to handle the v18 release this month
```
richardlau, line 37: The more pertinent point of the discussion was that Juan does not currently have a release key, and that even if he were to generate a new one it would not be something already in the README.

RafaelGSS (approved):

doc/meetings/2023-01-12.md
ruyadorno: oh that's a good point, so we should replace his name for someone else right away is what you're saying?

ruyadorno: updated to add this note to the minutes 😊

TROUBLE4UALL: Ok today is that day

TROUBLE4UALL: Eoc
Thank you for all the work that y'all put in I'm very happy with the end result so this should do it I should be able to deploy this contract with no problem please let me know something before I do this action

PR #833, opened 2023-03-14
doc: ncu-team sync update

BethGriggs:
`ncu-team sync README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,17 +223,7 @@ and CITGM team members listed below.
 
 ## CITGM team
 
-<!-- ncu-team-sync.team(nodejs/citgm) -->
-
-* [@al-k21](https://github.com/al-k21) - Oleksandr Kushchak
-* [@BridgeAR](https://github.com/BridgeAR) - Ruben Bridgewater
-* [@bzoz](https://github.com/bzoz) - Bartosz Sosnowski
-* [@gdams](https://github.com/gdams) - George Adams
-* [@MylesBorins](https://github.com/MylesBorins) - Myles Borins
-* [@richardlau](https://github.com/richardlau) - Richard Lau
-* [@targos](https://github.com/targos) - Michaël Zasso
-
-<!-- ncu-team-sync end -->
+* https://github.com/nodejs/citgm#citgm-team
 
 ## Emeritus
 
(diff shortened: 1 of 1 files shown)


ljharb (approved):

ljharb:
seems like maybe the release team isn't on the access list for this repo. should it be?

BethGriggs:
>  seems like maybe the release team isn't on the access list for this repo. should it be?

Oh hmm, it looks like CITGM GitHub team is not a subteam of the parent Release WG team. I cannot remember if that was always the case or the team structure got changed at some point.

Not sure whether we should make it a subteam, or just grant the CITGM team access to this repository also.

ljharb:
subteams get all the access their parent teams have, so it'd be more like, citgm was a parent of release - which probably doesn't make sense.

so i'd assume adding the citgm team specifically would make the most sense

richardlau:
Why would the CITGM team need access to the release repository?

ljharb:
@richardlau perhaps they don't. but this repo's readme lists the citgm team, and my approval here was gray instead of green.

¯\\\_(ツ)_/¯ just bringing it up, no strong feelings here

BethGriggs:
@ljharb it's a fair comment to raise - conceptually we have, and our wording implies, that the CITGM team is a subteam of the wider-Release WG and all other subteams of release have write access. If CITGM is an exception then we should be clear why.

> Why would the CITGM team need access to the release repository?

At least to maintain their list of members. It's reasonable to  decide not to grant access, but then maybe the member list for CITGM should be listed in a place where the team has access to review it.

Trott:
> @richardlau perhaps they don't. but this repo's readme lists the citgm team, and my approval here was gray instead of green.
> 
> ¯\_(ツ)_/¯ just bringing it up, no strong feelings here

CITGM was removed as a subteam from Release based on the conversations in https://github.com/nodejs/admin/issues/737. As such, the thing to do is probably to move the CITGM members list to the CITGM repository.

BethGriggs:
Updated based on https://github.com/nodejs/citgm/pull/944

richardlau (approved):

targos (approved):

Issue #547, opened 2020-03-09
Suggestion: drop backporters team?


BethGriggs:
At the moment, everyone in the `backporters` team is also in the `lts` team. I'm not sure the distinction is still necessary (often enough @nodejs/lts is notified in PRs).

I'd suggest reducing down to LTS, Releasers, and CITGM teams, thoughts?

targos:
sgtm

MylesBorins:
The one challenge I would see with this is that the LTS team has far more people on it than the `backporters` team and there is a single discrepancy @addaleax is on `backporters` but not `lts`.

We also use `backporters` for basic IAM, in that no one but that team can push to `vX.x-staging` branches.

So I would be in favor of this change but we would likely need to do the following

* remove inactive members for the `lts` team
* add @addaleax to the `lts` team
* update IAM for branch protection
* then remove backporters team

To be clear, I'm very much in favor of simplify team structure as well as pruning inactive folks!

sam-github:
I don't think backporters have ever landed on vx.y-staging branches, at least, I never have, don't know about @addaleax 

I didn't realize I was on nodejs/lts, I thought I was just on nodejs/backporters. I self-removed from both teams.

MylesBorins:
I've just double checked and it looks like the branch protection is not on the .x-staging branches after all, my mistake. I'm now recalling that we removed it because in the past it enforced `no force-pushes` which breaks our workflow.

This is now no longer and issue, but would be a good idea to enforce this policy after we sort out team membership

MylesBorins:
I've opened https://github.com/nodejs/Release/issues/549 to work on cleaning up the LTS list.

One thing that does strike me is that after removing backporters we won't really have a group for folks who want to participate in Release but not necessarily only work on LTS... it might make sense for us to completely rethink the structure of the group and how we use it for permissions. Right now we have

* Release
  - LTS (Only those in this group can do LTS releases)
    - Backporters (Only those in this group can land on v.x-staging for LTS)
  - Releasers (Only those in this group can do releases)
  - CITGM (Only those in this group can manage the CITGM repo)

Release itself has no members, only child members. Perhaps the current structure that we have doesn't make a ton of sense, since we have a mixture of IAM related roles and collaborative roles... and not a really great story for those who are ramping up or want to be more involved on a policy side. So here's a shot in the dark at what an alternative Structure could look like (names to be bikeshedded):

* Release (Top-Level group, no direct membership)
[...]

PR #912, opened 2023-09-15
update governance to reflect process

MylesBorins:
For some time all releasers have been allowed to make
LTS releases and backport changes. These documentation updates
are not intended to change any processes but rather to reflect
the current reality of how the team is operating.

There is still a backporters team that allows non-releasers to
backport commits, that is not currently documented.

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -164,7 +164,6 @@ reported to the TSC immediately so remedial action can be taken.**
 
 For releasing on LTS release lines, releasers must:
 
-* Be added as members of the GitHub [LTS team](https://github.com/orgs/nodejs/teams/lts)
 * Fully release at least one Current Node.js release; for practice, and to help promote the new GPG key to the user ecosystem prior to it being required to verify an LTS release
 
 ### SSH key guidance
--- a/README.md
+++ b/README.md
@@ -79,16 +79,13 @@ the working group does not automatically result in membership in these
 teams. These teams are:
 
 * Releasers team
-* LTS team
 * CITGM team
 
 The `releasers` team is entrusted with the secrets and CI access to be able
 build and sign releases. **Additions to the releasers team must be approved
-by the TSC following the process outlined in GOVERNANCE.md.**
-
-The Long Term Support (LTS) team manages the process/content of LTS releases
-and the required backporting for these releases. Additions to the LTS
-team needs sign off from the rest of the LTS team.
+by the TSC following the process outlined in GOVERNANCE.md.** The team also
+manages the process/content of LTS releases and the required backporting for
+these releases.
 
 The Canary in the Gold Mine (CITGM) team maintains CITGM as one of
 the key sanity checks for releases. This team maintains the CITGM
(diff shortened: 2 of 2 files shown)


GOVERNANCE.md
```diff
@@ -164,7 +164,6 @@ reported to the TSC immediately so remedial action can be taken.**
 
 For releasing on LTS release lines, releasers must:
 
-* Be added as members of the GitHub [LTS team](https://github.com/orgs/nodejs/teams/lts)
```
MylesBorins, line 167: This change might require TSC approval since it changes Governance.md

README.md
```diff
@@ -79,16 +79,13 @@ the working group does not automatically result in membership in these
 teams. These teams are:
 
 * Releasers team
-* LTS team
 * CITGM team
```
MylesBorins, line 82: I don't know if we should remove this as well or not... it seems like the maintenance of CITGM isn't just relasers anymore but not 100%

README.md
```diff
@@ -173,52 +170,20 @@ the discretion of the Release working group.
...
-* [@mhdawson](https://github.com/mhdawson) - Michael Dawson
-* [@MylesBorins](https://github.com/MylesBorins) - Myles Borins
-* [@RafaelGSS](https://github.com/RafaelGSS) - Rafael Gonzaga
-* [@richardlau](https://github.com/richardlau) - Richard Lau
-* [@ruyadorno](https://github.com/ruyadorno) - Ruy Adorno
-* [@targos](https://github.com/targos) - Michaël Zasso
-
-<!-- ncu-team-sync end -->
-
-### Backporters team
```
MylesBorins, line 194: we may want to bring this back into the readme, but there was no documentation about the backporters team

UlisesGascon (approved):

README.md
richardlau: Didn't we have branch protection rules to allow backporters to land things on the staging branches?

README.md
```diff
@@ -173,52 +170,20 @@ the discretion of the Release working group.
...
-
-<!-- ncu-team-sync end -->
-
 ## Releasers team
 
 <!-- ncu-team-sync.team(nodejs/releasers) -->
 
 * [@bengl](https://github.com/bengl) - Bryan English
-* [@BethGriggs](https://github.com/BethGriggs) - Bethany Nicolle Griggs
+* [@BethGriggs](https://github.com/BethGriggs)
```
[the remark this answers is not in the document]
richardlau: I think this is a result of `ncu-team sync` and probably related what is visible in @BethGriggs 's public profile.

README.md
```diff
@@ -79,16 +79,13 @@ the working group does not automatically result in membership in these
...
-* LTS team
 * CITGM team
 
 The `releasers` team is entrusted with the secrets and CI access to be able
 build and sign releases. **Additions to the releasers team must be approved
-by the TSC following the process outlined in GOVERNANCE.md.**
-
-The Long Term Support (LTS) team manages the process/content of LTS releases
-and the required backporting for these releases. Additions to the LTS
-team needs sign off from the rest of the LTS team.
```
richardlau, line 91: So this is an interesting reality vs ideal situation. For as long as I have been involved with the Release WG the LTS team has not been the team with permissions to do releases (even LTS releases) -- that's the releasers team. It is supposed to be the team that discusses/gets to decide LTS policy/process/content. It is not a requirement that all members of the Release WG become releasers.

It's a reasonable question as to whether we still need a LTS team distinct from the whole Release WG. I know I've pinged the LTS team in the past. 

Release, like a number of other Working Groups, has the challenge of dwindling participation over time punctured by occasional drives to increase membership.

GOVERNANCE.md
richardlau: I wonder if this dates back to when the group was the Long Term Support (LTS) group and this repository was called "lts" and before it was expanded to include current releases?

README.md
RafaelGSS: I agree. We discussed a not long time ago that the distinction between backporters/lts-team might not be necessary anymore since all the releasers fall in the same group.

BethGriggs: Some related conversations in https://github.com/nodejs/Release/issues/547.

PR #1090, opened 2025-05-06
add Krypton to version 24's codename

methbkts:


richardlau:
We don't add the codenames until the release transitions to Long Term Support (LTS).

PR #1108, opened 2025-06-26
Update schedule.json to reduce v25.x Maintenance

aduh95:
We've discussed in the Release WG meeting about how having long maintenance put a too heavy burden for security releases. For example, we did release Node.js 23.11.1 after Node.js 24.0.0, which takes probably more volunteer time that it brings value to the community.

I suggest we reduce the maintenance window, I picked 2026-05-06 because 24.0.0 was released on 2025-05-06 (so as long as 26.0.0 is not as delayed as 24.0.0 was, we don't have a gap where v25.x is EoL before v26.x is released).

wdyt?


aduh95:
Not really, not for odd-numbered release lines at least. There's https://github.com/nodejs/Release?tab=readme-ov-file#release-plan:

https://github.com/nodejs/Release/blob/cd5a754711646de14ac11adaea504b69e5a9b6bb/README.md?plain=1#L105

